PR #45, opened 2024-08-01
Add Server Group Cleanup

spjmurray:
Nova just orphans things on project deletion, which leaves a bunch of server groups hanging around in the aether, so we should play nice and clean these up for it, and make life easier on our platform teams.


spjmurray:
Cluster creation works fine, and deletion cleans up the server group.

PR #73, opened 2024-11-13
Add server provisioner

nsricardor:
This PR adds the server provisioner and corresponding logic on the openstack provider.

### DIFF
--- a/charts/region/crds/region.unikorn-cloud.org_servers.yaml
+++ b/charts/region/crds/region.unikorn-cloud.org_servers.yaml
@@ -23,6 +23,9 @@ spec:
     - jsonPath: .metadata.creationTimestamp
       name: age
       type: date
+    - jsonPath: .status.privateIP
+      name: privateIP
+      type: string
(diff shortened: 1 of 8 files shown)


spjmurray (commented):
Surprisingly small!

spjmurray:
Okay, so first issue was this doesn't actually work on all clouds as (my fault) I allowed myself to get pressurized into supporting physical VLANs as a priority and not thinking about the bigger picture... but with that in place.

Test 1 (non-VLAN environment):

* Create identity
* Create security group and ingress rule for 22 (https://github.com/unikorn-cloud/region/issues/74 opened as a follow up)
* Create network
* Create server

```
$ ssh -i id.rsa -l ubuntu 193.16.42.107 hostname
foo
```

* Delete identity and everything cascades (https://github.com/unikorn-cloud/region/issues/75 opened as a follow up)

Good start!

spjmurray (approved):
Good enough for me!

PR #79, opened 2024-11-27
Return server status

nsricardor:
This PR addresses a small fix to the server API endpoints by populating and returning the `status` field in the server response. Specifically, it ensures that the status field includes both the `public and private IP` addresses of the provisioned server.

The status field, which is already specified in the `server.spec.yaml` file, will now contain these IPs, allowing client services like Compute and Slonk to be aware of the server's provisioned network interfaces.


spjmurray (approved):
May be useful for an external observer 😸

PR #85, opened 2024-12-03
Add server userdata to support server configuration and init scripts

nsricardor:
Add support for the `userdata` property to provide custom initialization data when a server is first launched. This allows for automation of tasks such as configuring the server, installing software, or setting environment-specific settings, all without requiring manual input.


spjmurray (approved):
Seems agreeable!  I'll merge it as we're working on testing/UI integration in parallel!

spjmurray:
Actually, do be a pal and bump the `charts/unikorn-region/Chart.yaml` version to rc2.  Won't be much good if you cannot import the types!

PR #90, opened 2024-12-10
Drop image selector

nsricardor:
This PR removes the image selector feature from the region service to keep the region more generic and independent of any image selection logic.


spjmurray (approved):
Looks good!

PR #105, opened 2025-02-28
Add support for configuring Port Allowed Address Pairs

nsricardor:
This PR allows users to define additional IP ranges that can be routed through an OpenStack Port. It enables servers to function as NAT/VPN gateways by configuring the necessary Allowed Address Pairs, ensuring that traffic is not blocked. This change is essential when the network topology includes a server functioning as a VPN gateway (e.g., WireGuard), as OpenStack would otherwise block routed traffic without this configuration.


spjmurray:
Remind me to block all phone notifications until 9am 😸

spjmurray:
I'm guessing if it was unidirectional traffic from outside to inside, this would all be handled by a SNAT, however the problem is stuff from inside going outside needs to send to an external IP via a server acting as a router, which it's blocking.  I seem to remember doing this stuff like a decade ago, brain rot etc.

nsricardor:
> I'm guessing if it was unidirectional traffic from outside to inside, this would all be handled by a SNAT, however the problem is stuff from inside going outside needs to send to an external IP via a server acting as a router, which it's blocking. I seem to remember doing this stuff like a decade ago, brain rot etc.

Yep, that's the case. One server in the cluster is running WireGuard and acts as a router, directing traffic from the other nodes to an external network.

spjmurray (commented):
Have at it!

spjmurray (approved):

PR #108, opened 2025-03-12
Allow Unlimited Snapshots

spjmurray:
We really really need to go back and revisit the quota stuff or we're leaving the entire system at risk of a denial of service attack by some errant workload...


spjmurray:
Note: tested the upgrade does actually retroactively apply the new defaults.

PR #113, opened 2025-03-27
Automate Image Selection

### LINKED ISSUE #112: Filter Images Based on Schema
We have a specification, so reject any images that don't meet the criteria.

https://github.com/unikorn-cloud/specifications/blob/main/specifications/providers/openstack/flavors_and_images.md#image-properties

There is also a PoC sanitizer in https://github.com/spjmurray/unikorn-images/ so you can essentially just lift and shift the code!

spjmurray:
Prior to this change, we let the user select what images were presented to the user.  Since the time when this was written, we have released a versioned specification that controls what consitutes a usable image, taking into account UX requirements, and those of various higher level services e.g. Kubernetes.  Remove the capability for a user to manaually select images and instead use the schema defined by the specification. The signing key is still available to verify the provenance of an image on platforms where users can, legitmately, create a valid image compatible with the schema.

Fixes #112


charts/region/values.schema.json
```diff
@@ -313,12 +313,6 @@
                   "selector": {
                     "type": "object",
                     "properties": {
-                      "properties": {
```
spjmurray, line 316: Note: this is a breaking change, and will need a release note.

pkg/providers/openstack/image.go
```diff
@@ -32,10 +32,10 @@ import (
 	"github.com/gophercloud/gophercloud/v2"
 	"github.com/gophercloud/gophercloud/v2/openstack"
 	"github.com/gophercloud/gophercloud/v2/openstack/image/v2/images"
+	"github.com/kaptinlin/jsonschema"
```
spjmurray, line 35: Technically there is a kube-openapi package, to keep things in the "more trustworthy" ecosystem, but this library is way simpler and not over-engineered like everything Kubernetes!

pkg/providers/openstack/image.go
```diff
@@ -154,20 +215,22 @@ func (c *ImageClient) verifyImage(image *images.Image) bool {
 	return ecdsa.VerifyASN1(signingKey, hash[:], signature)
 }
 
-func (c *ImageClient) filterImage(image *images.Image) bool {
+// imageValid returns true when the image is active, matches the schema and optionally
+// is signed by a trusted image building pipeline.
+func (c *ImageClient) imageValid(image *images.Image, schema *jsonschema.Schema) bool {
```
spjmurray, line 220: Changing the name and logic to make it easier to understand.

spjmurray:
Go on Michael, let's see how evil you are 😸

spjmurray:
Oh yeah, testing notes.  Applied locally with no drama, can still see images in the UI for both compute and k8s services.

pkg/providers/openstack/image.go
```diff
@@ -52,6 +52,81 @@ var (
 	ErrKeyType = errors.New("key type unsupported")
 )
 
+// imagePropertySchemaV2 defines what consitutes a valid image e.g. contains all the
+// required information to work correctly.  This is defined in:
+// https://github.com/unikorn-cloud/specifications/blob/main/specifications/providers/openstack/flavors_and_images.md.
+const imagePropertySchemaV2 = `{
```
squaremo, line 58: Might be slightly neater to put this in a file and use https://pkg.go.dev/embed. (You wouldn't have to cast to []byte later, too)

charts/region/values.schema.json
squaremo: What intervention would be needed to upgrade a running system past the change?
Could this be made backward-compatible by keeping it in, but deprecating its use?

squaremo (changes requested):
OK, I think I understand this change. In my own words: instead of supplying tags ("properties") when specifying images that can be used in an OpenStack region spec, filter to any image that has the required labels according to the schema (and if key is supplied, then signed by that key).

Suggested changes:

 - I think it's worth having some unit tests that demonstrate `imageValid()` works (and continues to work) the way expected. This will cover off testing that the schema is correct, as well.
 - Is it possible to just deprecate `ImageSelector.Properties`, so it doesn't need any intervention to upgrade?

spjmurray:
Oh you are one of them eh?  😉  Guess I have some cycles to add in the units and battle with crypto, brb...

spjmurray:
Put that in your pipe and smoke it!

charts/region/values.schema.json
spjmurray: Well, we're going to have to remove it eventually, I see no difference doing it now compared until later 🤷🏻  If you are insistent I may be persuaded with a tequila.

pkg/providers/openstack/image_test.go
```diff
@@ -0,0 +1,289 @@
...
+
+		return image
+	}
+}
+
+func gpuImageFixture() *images.Image {
+	image := basicImageFixture()
+	image.Properties[gpuVendorProperty] = gpuVendorNVIDIA
+	image.Properties[gpuModelsProperty] = "H200"
+	image.Properties[gpuDriverVersionProperty] = "Lewis Hamilton (Ferrari Edition)"
```
spjmurray, line 97: 💯  for this Easter egg alone?

squaremo: 💯 :-D

pkg/providers/openstack/image_test.go
squaremo, line 1: Good tests, nice one 💯 
In particular, easy to read and lots of bases covered, efficiently.

squaremo (approved):

Issue #103, opened 2025-02-20
Create a Kubernetes Region Provider


spjmurray:
For once something simple!  The goal here is to allow users to provision Kubernetes clusters in their providers of choice and allow abstraction and aggregation in the region service.

* API simply reports:
  * I'm Kubernetes
  * Here's my `kubeconfig`
  * Probably want to describe the underlying cluster to a certain degree:
    * How big is it for scheduling decisions
    * What GPUs are in it?
* Refuses to create identities, servers or any of that jazz

PR #115, opened 2025-04-10
K8S Lite

spjmurray:
Adds a kubernetes region type, provides a way to get access to the kubeconfig from the Kubernetes controller, creates a method to allow "flavors" to be derived from the cluster that's handed to us.

Implements #103


spjmurray:
Note: I'm breaking the rules here due to time constraints and I need the new openapi client and types to link against to make progress elsewhere...

PR #155, opened 2025-07-01
Principal Propagation

spjmurray:
As a leaf service, this simply updates to the new principal aware identity interfaces so it's propagated down to the Kubernetes resources. This can be used to add properties to VMs and bare metal instances which can then be polled for billing purposes, for those who like to do it from the bottom, rather than just look at the cluster.  This hasn't been done yet to limit initial scope.

The client however, that will be consumed by the compute service has been extended to allow additional mutation functions, specifically principal header injection.


spjmurray:
This is literally easy now, you've already reviewed all the changes in Identity 🎉

PR #166, opened 2025-09-02
Add Server Power Lifecycle Management APIs

alansy-nscale:
## Description
This PR introduces a new set of APIs for managing server power lifecycles, including starting, stopping, and performing both soft and hard reboots. In addition, the region monitor now tracks server power state changes alongside server health.


alansy-nscale:
Yes, that is expected, as I did not modify the phase of the machine before rebooting it. Should we change the state to `Pending`?

spjmurray:
> Yes, that is expected, as I did not modify the phase of the machine before rebooting it. Should we change the state to `Pending`?

I would probably expect to see a `rebooting` message as an end user.  Consult with the product team!

alansy-nscale:
> I would probably expect to see a rebooting message as an end user. Consult with the product team!

I looked into AWS and GCP, and both keep the instance in the `Running` state during a reboot, but I am happy to add a new `Rebooting` state if we decide we want that.

- https://cloud.google.com/compute/docs/instances/instance-lifecycle
- https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html

spjmurray (approved):

PR #167, opened 2025-09-04
Extract Lifecycle Phase Enum into Shared Schema

alansy-nscale:
## Description
This PR includes changes that extract the `serverStatus.phase` enum values into a dedicated schema called `InstanceLifecyclePhase`. The `serverStatus.phase` property now references this shared schema, allowing the `InstanceLifecyclePhase` definition to be reused in other specifications. It also includes a bug fix for the reboot handlers, where the functions were swapped, causing the hard reboot handler to trigger a soft reboot and the soft reboot handler to trigger a hard reboot.

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -537,7 +537,7 @@ func (h *Handler) PostApiV1OrganizationsOrganizationIDProjectsProjectIDIdentitie
 		return
 	}
 
-	if err := server.NewClient(h.client, h.namespace).SoftReboot(r.Context(), organizationID, projectID, identityID, serverID); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/handler/handler.go
```diff
@@ -537,7 +537,7 @@ func (h *Handler) PostApiV1OrganizationsOrganizationIDProjectsProjectIDIdentitie
 		return
 	}
 
-	if err := server.NewClient(h.client, h.namespace).SoftReboot(r.Context(), organizationID, projectID, identityID, serverID); err != nil {
+	if err := server.NewClient(h.client, h.namespace).Reboot(r.Context(), organizationID, projectID, identityID, serverID, true); err != nil {
```
spjmurray, line 540: It seems to me this is less readable now, as you have a magic boolean, but it's not that big of a gripe.

PR #170, opened 2025-09-18
Return Images Relevant to the Requester

alansy-nscale:
## Description
The list images endpoint now filters results based on the new attributes in `image.properties` to return images relevant to the requester.
- If the attribute `unikorn:organization:id` is absent, the image is treated as public and accessible to all users.
- If the attribute `unikorn:organization:id` is present, the image is only returned when the requester's ACL organization ID matches the image's organization ID.

### DIFF
--- a/pkg/providers/openstack/provider.go
+++ b/pkg/providers/openstack/provider.go
@@ -351,7 +351,7 @@ func (p *Provider) Flavors(ctx context.Context) (types.FlavorList, error) {
 }
 
 // Images lists all available images.
-func (p *Provider) Images(ctx context.Context) (types.ImageList, error) {
+func (p *Provider) Images(ctx context.Context, organizationID string) (types.ImageList, error) {
 	imageService, err := p.image(ctx)
 	if err != nil {
 		return nil, err
@@ -361,6 +361,12 @@ func (p *Provider) Images(ctx context.Context) (types.ImageList, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	// Filter images that either have no organisation ID or match the requested one.
+	resources = slices.DeleteFunc(resources, func(image images.Image) bool {
+		value := image.Properties["unikorn:organization:id"]
+		return value != nil && value != organizationID
+	})
 
 	result := make(types.ImageList, len(resources))
 
--- a/pkg/handler/region/region.go
+++ b/pkg/handler/region/region.go
@@ -359,7 +359,7 @@ func (c *Client) ListImages(ctx context.Context, organizationID, regionID string
 		return nil, errors.OAuth2ServerError("failed to create region provider").WithError(err)
 	}
 
-	result, err := provider.Images(ctx)
+	result, err := provider.Images(ctx, organizationID)
 	if err != nil {
 		return nil, errors.OAuth2ServerError("failed to list images").WithError(err)
 	}
--- a/pkg/providers/kubernetes/provider.go
+++ b/pkg/providers/kubernetes/provider.go
@@ -194,7 +194,7 @@ func (p *Provider) Flavors(ctx context.Context) (types.FlavorList, error) {
 }
 
 // Images lists all available images.
-func (p *Provider) Images(ctx context.Context) (types.ImageList, error) {
+func (p *Provider) Images(ctx context.Context, organizationID string) (types.ImageList, error) {
 	return nil, ErrUnimplmented
 }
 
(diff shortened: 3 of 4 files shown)


pkg/handler/server/client.go
```diff
@@ -111,6 +112,29 @@ func (c *Client) List(ctx context.Context, organizationID string, params openapi
 
 // Create instantiates a new resource.
 func (c *Client) Create(ctx context.Context, organizationID, projectID, identityID string, request *openapi.ServerWrite) (*openapi.ServerRead, error) {
+	identity, err := identity.New(c.client, c.namespace).GetRaw(ctx, organizationID, projectID, identityID)
```
spjmurray, line 115: Region is part of the request spec surely?

pkg/handler/server/client.go
```diff
@@ -111,6 +112,29 @@ func (c *Client) List(ctx context.Context, organizationID string, params openapi
...
+	if err != nil {
+		return nil, err
+	}
+
+	provider, err := region.NewClient(c.client, c.namespace).Provider(ctx, identity.Labels[constants.RegionLabel])
```
spjmurray, line 120: Surely you really want to add a Get interface to the region API rather than filtering?

pkg/handler/server/client.go
alansy-nscale: Do you mean the `regionID`?

pkg/handler/server/client.go
alansy-nscale: Do we want to add a get image interface to the `ImageClient`? The `ImageClient` currently has a `TimeoutCache[[]image.Image]` for images retrieved from the list endpoint. If we add a new get interface, should the image be cached?

pkg/providers/openstack/image.go
```diff
@@ -197,36 +206,35 @@ func (c *ImageClient) images(ctx context.Context) ([]images.Image, error) {
...
 }
 
-func ImageSchema() (*jsonschema.Schema, error) {
-	return jsonschema.NewCompiler().Compile(imagePropertySchemaV2)
-}
-
-// Images returns a list of images.
-func (c *ImageClient) Images(ctx context.Context) ([]images.Image, error) {
-	result, err := c.images(ctx)
+func (c *ImageClient) GetImage(ctx context.Context, id string) (*images.Image, error) {
```
alansy-nscale, line 224: I'm not sure if we need this. We can just handle the filtering in the provider layer.

pkg/handler/server/client.go
```diff
@@ -137,6 +157,27 @@ func (c *Client) Get(ctx context.Context, organizationID, projectID, serverID st
...
+	if err != nil {
+		return nil, errors.OAuth2ServerError("failed to create region provider").WithError(err)
+	}
+
+	// REVIEW_ME: Is the `unikorn:organization:id` property immutable on the image?
```
spjmurray, line 170: For now yes, I don't see a world where you'd transfer between orgs.  You can always re-upload!

pkg/providers/openstack/provider.go
```diff
@@ -464,6 +401,104 @@ func (p *Provider) imagePackages(image *images.Image) *types.ImagePackages {
 	return &result
 }
 
+func (p *Provider) isImageBelongsToOrganization(image *images.Image, organizationID string) bool {
+	value := image.Properties["unikorn:organization:id"]
+	return value != nil && value != organizationID
```
spjmurray, line 406: Given how this is being used (removing images private to this organization), I'd prefer something like `imagePublicOrOwnedByOrganization()`.  This logic would then be:

`return value == nil || value == organizationID`

Which is easier to understand I think 😄  Especially seeing as the function name and logic are at odds with one another!

pkg/providers/openstack/provider.go
```diff
@@ -464,6 +401,104 @@ func (p *Provider) imagePackages(image *images.Image) *types.ImagePackages {
...
+	}
+
+	resources, err := imageService.ListImages(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// Filter images that either have no organisation ID or match the requested one.
+	resources = slices.DeleteFunc(resources, func(image images.Image) bool {
+		return p.isImageBelongsToOrganization(&image, organizationID)
```
spjmurray, line 466: See above, this would be `return !imagePublicOrOwnedByOrganization(&image, organizationID)`

pkg/providers/openstack/image.go
spjmurray: Yeah, I say we do it in the provider, it's cached anyway.

spjmurray:
We really should back this up with unit tests to prove it works and also stop it from breaking... problem is I don't think this is a well engineered bit of code to pull that off.  Let me sleep on it 💤

PR #187, opened 2025-10-10
Add References API to Network Resources

spjmurray:
The file storage service is being migrated out of tree and thus will no longer be allowed to directly manipulate network resources as that would breach a security boundary.  This introduces a references API and client that allows the file storage service to add and remove finalizers from a network resource as required.

This follows the precedent set by the identity service and its project references API thus maintaining consistency across all components.


nsricardor (dismissed):
Nice one! this does exactly what it sets out to do and stays consistent with the uni-identity approach for managing project references. Happy to adopt this in uni-storage once released.

spjmurray:
Sorry, got distracted by something I never quite understood... why do we have to URL encode the references when the client should be doing it for us?  Turns out the server code actually decodes twice, one in the chi router and again in the OAPI router.  Codex has agreed this is the correct, albeit imperfect fix for now!

nsricardor (approved):

PR #364, opened 2026-01-28
Factor out remaining handlers

squaremo:
This PR continues the work started in recent server and image API changes, of separating the implementations of sets of related APIs into their own structs with methods.

The methods are broken down along the same lines as the clients (subpackages of pkg/handler) are. This means they can in general depend on one of those packages.

This accomplishes a few nice to have things:

 - It localises and deduplicates the code that constructs a client, for each handler, which makes it less trouble to change details (e.g., so as to provide extra arguments)
 - It means each handler can have tests written just for that handler, cutting down the dependencies that need to be provided. (With mock generation this is a less serious problem in practice, here at least; but still a good habit)
 - It normalises how handlers work (so it's clear what to copy and paste for a new one :-))
 - it breaks the handlers down into smaller files which each have a coherent set of definitions in them, which makes it easier to navigate

This paves the way to further normalisation, and in particular, makes it easier to introduce an abstraction for obtaining providers per https://github.com/nscaledev/uni-region/pull/335.

The work is largely mechanical refactoring, as evidenced by the minimal changes out side this package (and to tests). Complications are noted in the commit messages.


pkg/handler/server/client.go
```diff
@@ -35,25 +35,51 @@ import (
...
+	"github.com/unikorn-cloud/region/pkg/providers"
+	"github.com/unikorn-cloud/region/pkg/providers/types"
 
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// Provider gloms together the operations needed to provision and control servers.
```
squaremo: Moved as part of combining server v1 and server v2 clients.

PR #409, opened 2026-02-13
Provider Startup And Image Cache Warmup

spjmurray:
This stack now does three related things to make provider startup and image caching behave predictably in production and in tests.

1. Create shared providers at process startup without depending on a cold controller-runtime cache.
The provider bootstrap path now separates the bootstrap reader from the retained runtime client so controllers can initialize shared provider state before the manager cache starts. This keeps the fail-fast startup behaviour we want while avoiding the pre-start cache read failure that showed up in integration. The same startup path now also initializes the monitor's shared Providers value once, instead of rebuilding providers during every health check.

2. Keep the image super cache warmup behaviour, but make that policy explicit.
The old boolean cache switch is replaced with explicit warm-image-cache options so it is obvious why the API server warms image data before becoming ready while other call sites do not.

3. Make startup preloading and post-startup fixture/region creation compose cleanly.
Providers are still preloaded eagerly for readiness, but a lookup miss now lazily instantiates and caches a provider for a newly created Region. That fixes the integration case where regions created after process startup were invisible until rollout.

Because those lazy provider lookups now use direct get calls for Region objects, the chart RBAC also needs to grant get on regions to the API, monitor, and controller service accounts. The API also needs get on filestorageclasses for its direct storage class reads.

4. Fix the POST /api/v2/networks not-found path so it returns a real 404.
The stack now declares the matching 404 in the OpenAPI schema for POST /api/v2/networks. That is an explicit bug fix: a legitimate region-not-found path is now surfaced cleanly as a 404 instead of panicking the validator and degenerating into an nginx 502.

The readiness probe commit then ensures the API is only marked ready once that startup initialization has completed, which is the behaviour we want for zero-downtime rollouts.

Review context updates:
- The current branch keeps `context.Context` on the provider image query interface; older review comments about removing it refer to an intermediate revision, not current HEAD.
- The generic provisioner/provider reuse follow-up discussed in review landed separately in #421, so those comments are not pending changes for this PR.


pkg/provisioners/internal/base/base.go
```diff
@@ -50,7 +50,14 @@ func Provider(ctx context.Context, object client.Object) (types.Provider, error)
 		return nil, err
 	}
 
-	return providers.New(cli, object.GetNamespace()).LookupCloud(ctx, object.GetLabels()[constants.RegionLabel])
+	// TODO: pre-allocate this and stash it in the context for persistence across
+	// reconciles.
```
squaremo: Please don't put this in a context. It makes it dynamically scoped, so harder to reason about (for people and in testing); and in this case specifically, it's brittle because it depends on the same context being reused later (which is definitely not how one is supposed to use contexts).

pkg/providers/types/interfaces.go
```diff
@@ -26,7 +26,7 @@ import (
 
 type ImageQuery interface {
 	// List runs the query and returns the result
-	List(ctx context.Context) (ImageList, error)
+	List() (ImageList, error)
```
squaremo: I think it's worth keeping the context in the arguments.
 - changing the interface is not necessary for this PR, and it adds a bunch of noise in the diff
 - having a context arg keeps it consistent with other "fetch" methods, even if it's not used in the implementation
 - it could reasonably be used in an implementation (a new provider, or just if we could be a little more clever with the OpenStack provider), in which case it'd have to be changed back

pkg/providers/providers.go
```diff
@@ -58,41 +57,62 @@ func ProviderToServerError(err error) error {
...
-	return &providersImpl{
-		client:    client,
-		namespace: namespace,
-		cache:     map[string]types.CommonProvider{},
+// New creates and synchronously initializes all region providers.  The chosen behaviour
+// is to fail on intialization failure, the reasoning is that during upgrade an old version
+// of the service should be running to handle traffic, and we'd rather not have something
+// put into production that is known to be broken.  On that note, regions aren't lazily
+// loaded any more, so you will need to manually rollout the deployment to pick them up.
+func New(ctx context.Context, c client.Client, namespace string, withCaches bool) (Providers, error) {
```
squaremo: This is an awkward leak through the abstraction. Caches are supposed to be hidden in the implementation. This is going back to the provider-but-some-methods-won't-work situation.

On the other hand I do agree it's not great if the provisioners each have image caches for each OpenStack region, needlessly spamming the OpenStack API.

How about this:

What we want is to give the provisioners something like Providers, but you don't get the full implementations.

* Adding an interface, say `ProvisionerProvider`, and a `providers.LookupProvisioner(...)` method to providers does not suffice, because it'll still be initialised to the full implementation including the cache.
* So it needs a different set of provider constructors (`.../internal/openstack.NewForProvisioner`?), and different methods (you can _only_ call `LookupProvisioner()`) which means a different Providers type and constructor, though it may be able to share some bits.

I don't really like the extra, parallel stuff, but I don't really like the leaky abstraction either. Maybe this'll spark a better idea.

squaremo (commented):
Most of this I can live with, because it doesn't change semantics, and it can be reworked without changing the behaviour again. (That's all the inconclusive comments around provisioners and how they get providers)

The one thing that is a behaviour change is that this initialises providers on startup, rather than when they are requested. This is necessitated by wanting to warm the caches before being regarded as ready, which is sensible. It does mean that if there's a _new_ region, it won't be noticed. We should have a mitigation in place for that, even if it's documenting that regions won't be noticed until there's a controller restart.

I'm making this a comment rather than an approve, in case you want to act on the above, or any of the comments, before it's sealed. (Oh and it depends on an unmerged PR in core)

spjmurray:
On the main point, we could quite easily have a watch on Regions that triggers addition/deletion.  The providers themselves are responsible for updates to things like credentials etc.

pkg/providers/internal/openstack/provider.go
```diff
@@ -695,70 +704,66 @@ func convertImage(image *images.Image) (*types.Image, error) {
 	return &providerImage, nil
 }
 
-type imagePredicate func(*images.Image) bool
+type imagePredicate func(*types.Image) bool
 
+// TODO: this operates on generic types now, so shouldn't live here.
```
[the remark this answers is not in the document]
squaremo: > But it's generic to the common intermediate Image type, not openstack specific

Sure, but that's incidental. It's only _used_ by the OpenStack implementation. It's moot whether any other provider implementation that comes along would need it.

> all RBAC is typically done in the handler, with the exception of this stuff.

Are you thinking specifically of the check on organisation ID? That's a query parameter, not RBAC. It's like only returning the Kubernetes objects that have a particular organisation ID in a label (like, say, pkg/handler/network/client_v2.go#L104 or thereabouts); except, images aren't recorded in Kubernetes, they are in OpenStack, so that query has to be propagated through to the provider API.

pkg/providers/internal/openstack/provider.go
```diff
@@ -695,70 +704,66 @@ func convertImage(image *images.Image) (*types.Image, error) {
...
+			return false
 		}
-
-		im, err := convertImage(&images[i])
-		if err != nil {
-			return nil, err
-		}
-
-		result = append(result, *im)
 	}
```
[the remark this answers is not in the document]
spjmurray: Let's think... in python that would be

```python
any(f(i) for f in filters)
```

Which would translate roughly to...  I'm not going to bother it's too horrid!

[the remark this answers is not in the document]
squaremo: It works out nicely if the predicates are used as filters (which is, ultimately, what DeleteFunc wants anyway). I can submit as a follow-up.

squaremo:
> * Fail-fast philosophy: "we'd rather not have something put into production that is known to be broken"

The problem with this is that it's not what you're putting into production, it's what is there already. Say a release with these changes is running in production, happily. Then someone adds a duff region definition. The service will truck merrily along, until a new release is rolled out, _then_ it will refuse to start.

This _is_ an improvement, because it unmistakably surfaces a problem -- the duff region -- it just happens remotely to the problematic change. (And it doesn't support the claimed benefit quoted above.)

spjmurray:
I have updated the PR body to remove stale review context and make the current state explicit.

Exactly what was fixed in the PR message:
- Added an explicit note that the current branch still keeps `context.Context` on the provider image query interface, so older review comments about removing it refer to an intermediate revision rather than current HEAD.
- Added an explicit note that the generic provisioner/provider reuse follow-up discussed in review landed separately in #421, so those comments are not pending work for this PR.
- Kept the RBAC explanation aligned with the current code: lazy provider loading requires direct `get` on `Region` objects, while API `get` on `FileStorageClass` is for direct storage-class reads rather than provider lazy loading.

This is intended to stop automated review from re-litigating outdated comments from earlier revisions.

pkg/providers/providers.go
squaremo: I had a sketchy go at this in #428, but it can be done post-hoc (or just left alone).

squaremo:
> but a lookup miss now lazily instantiates and caches a provider for a newly created Region. That fixes the integration case where regions created after process startup were invisible until rollout.

This means there's a one-off latency cost when a newly-introduced Region definition is accessed as a provider by the API server. The first request would have the long latency we're trying to avoid with this whole big change. But that's one request per new region introduced per API server replica -- not the end of the world. The caching elsewise is certainly an improvement.

squaremo:
>  API get on FileStorageClass is for direct storage-class reads rather than provider lazy loading.

I don't quite get this -- did this break as part of these changes, or was it already broken?

pkg/managers/providers.go
```diff
@@ -38,7 +38,11 @@ type ProvidersInit struct {
 var _ coremanager.ControllerInitializer = &ProvidersInit{}
 
 func (f *ProvidersInit) Initialize(ctx context.Context, mgr manager.Manager, opts *options.Options) error {
-	providers := providers.New(mgr.GetClient(), opts.Namespace)
+	providers, err := providers.New(ctx, mgr.GetAPIReader(), mgr.GetClient(), opts.Namespace, providers.Options{})
```
squaremo, line 41: Ack, yep otherwise you need to wait for the caches. This has bitten me often enough you think I would have learned.

pkg/providers/providers.go
```diff
@@ -88,14 +126,14 @@ func newProvider(ctx context.Context, client client.Client, region *unikornv1.Re
...
-	return p.lookup(ctx, regionID)
+func (p *providersImpl) LookupCommon(regionID string) (types.CommonProvider, error) {
+	return p.lookup(context.Background(), regionID)
 }
 
 // LookupCloud returns a provider as identified by the region ID and must be
 // a cloud type.
-func (p *providersImpl) LookupCloud(ctx context.Context, regionID string) (types.Provider, error) {
-	provider, err := p.lookup(ctx, regionID)
+func (p *providersImpl) LookupCloud(regionID string) (types.Provider, error) {
```
squaremo, line 135: I don't understand why it's considered better to use `context.Background()` rather than passing a context in.

squaremo (approved):
I'm going to approve this on the basis that I've seen most of it several times, the new bit looks OK, and it fixes a well-understood problem. (And the machine review does not appear unhappy).

pkg/providers/providers.go
spjmurray: • The likely reason was to separate two concerns that had been coupled before:

  - request handling and cancellation
  - provider cache initialization and lifetime

  Before the refactor, a cache miss in LookupCloud used the caller’s ctx, so provider resolution was
  effectively request-scoped even though the result was stored in a long-lived shared cache. That blurred
  ownership: a canceled request could abort work whose real purpose was to populate shared process state.

  After the refactor, startup initialization and lazy backfill became provider-system responsibilities
  instead. Existing regions are loaded during Initialize, and late-discovered regions are fetched
  directly on demand without depending on informer readiness. In that model, using context.Background()
  is consistent with the idea that provider lookup is no longer part of the request lifecycle, but part
  of maintaining the provider cache.

  Executive summary: this was most likely done to avoid informer-cache timing issues during
  initialization and to make provider resolution a cache-management concern rather than a request-scoped
  concern. The tradeoff is that lazy lookups no longer respect request cancellation, but that appears to
  have been accepted in exchange for cleaner separation of concerns and more reliable provider
  availability.

pkg/managers/providers.go
spjmurray: Upgrade hooks are a prime example!  it usually passes after a _few_ pod restarts 😄

pkg/providers/providers.go
squaremo: > The likely reason

Whaddya mean "the likely reason", you submitted this code 🤨

spjmurray:
> I don't quite get this -- did this break as part of these changes, or was it already broken?

Neither do I, neither does the author, I've done a deep dive and 🤷🏻  happy to let it slide in lieu of a more intensive RBAC audit.

PR #415, opened 2026-02-17
Add default reservations for all networks

spjmurray:
Networks without explicit reservations now always receive a /25 reservation with a /28 provider carve-out, regardless of when they were created.

The previous approach used a per-resource annotation to distinguish legacy networks (which inherited the /25+/28 implicit defaults) from newly created networks (which were treated as having no reservations). That distinction is removed: the same defaults apply uniformly, so the annotation, the V2 defaults constant, and the `UseNewReservationDefaults` helper are all gone. Explicit `spec.reservations` always win.

The separate `networkV2UpdateSpec` OpenAPI schema is also removed: the PUT request spec now references `networkV2Spec` directly, matching the GET response and making read-modify-write straightforward. This also eliminates the JSON round-trip in `convertCreateToUpdateRequest`, replaced by a direct struct copy the compiler can verify.

The nil guards in `dhcpRange` and `storageRange` are removed as dead code: both functions are only called via `EffectiveReservations()`, which never returns nil.

OpenAPI descriptions updated to reflect that omitting `reservations` on create always applies the platform default (/25 + /28), not "no reservations".


spjmurray:
# Test Notes

Coverage should be good enough via unit testing, but manually... I started with:

```
| allocation_pools     | 192.168.0.200-192.168.0.254
```

* On upgrade, reverts back to the full range as expected and no storage can be used:

```
| allocation_pools     | 192.168.0.2-192.168.0.254                    |
```

```
  openstack:
    networkID: 7e384e0a-b17e-45e3-84dc-31c0c9a00539
    subnetID: 6223b368-ea91-4f7e-aa91-a3285f05e80c
```

* With a prefix length, this is taken from the DHCP pool and made available for storage:

```
spec:
  prefix: 192.168.0.0/24
  reservations:
    prefixLength: 25
```

```
| allocation_pools     | 192.168.0.128-192.168.0.254                  |
```

```
  openstack:
    networkID: 7e384e0a-b17e-45e3-84dc-31c0c9a00539
    storageRange:
      endIP: 192.168.0.127
      startIP: 192.168.0.2
    subnetID: 6223b368-ea91-4f7e-aa91-a3285f05e80c
```

* With a prefix length and a VIP pool prefix, the storage range is reduced, DHCP stays as it is

```
spec:
  prefix: 192.168.0.0/24
  reservations:
    prefixLength: 25
    vipPoolPrefixLength: 28
```

```
| allocation_pools     | 192.168.0.128-192.168.0.254
```

```
  openstack:
    networkID: 7e384e0a-b17e-45e3-84dc-31c0c9a00539
    storageRange:
      endIP: 192.168.0.127
      startIP: 192.168.0.16
    subnetID: 6223b368-ea91-4f7e-aa91-a3285f05e80c
```

* Finally with the VIP pool covering the entire allocation, no storage is possible and the DHCP is updated to not include the VIPs

```
spec:
  prefix: 192.168.0.0/24
  reservations:
    prefixLength: 28
    vipPoolPrefixLength: 28
```

```
| allocation_pools     | 192.168.0.16-192.168.0.254                   |
```

```
  openstack:
    networkID: 7e384e0a-b17e-45e3-84dc-31c0c9a00539
    subnetID: 6223b368-ea91-4f7e-aa91-a3285f05e80c
```

squaremo (dismissed):
Not much to add to the machine's review.

- A couple of typos, one in API docs
- Happy to see tests included
- Less happy to have one big diff in a single commit

pkg/openapi/server.spec.yaml
```diff
@@ -2126,6 +2128,44 @@ components:
...
+            are not available for end user use yielding N-2 usable
+            addresses.  This field is immutable so ensure your reservation
+            is large enough for potential use during the lifetime of the
+            network.  The reservation prefix length must be a positive
+            integer with a maximum of 30 and must be greater than the network
+            prefix length.
+          type: integer
+          minimum: 0
+          maximum: 30
+        vipPoolPrefixLength:
```
nsricardor: I have some concerns about the use of vipPoolPrefixLength here. “VIP pool” is a fairly generic term and is already used by VAST (and potentially other network services). Reusing the term it in this context could create confusion for users and future maintainers, particularly when troubleshooting or discussing integrations.

squaremo (commented):
I have a couple of concerns that stop me green-lighting this:

1. `nonStoragePrefixLength` seems like an unfriendly term to freeze in an API. I get Ricardo's objection to "vip pool"; is the distinction to draw that of reserved for the system vs reserved for the user?

2. So far as I can tell, the reservation is supplied in an update request (`networkV2Update` includes `networkv2Spec`), but ignored (the update code passes `current.Spec.Reservations`). If it's immutable, I would expect it to not be in the request, or to provoke an error if it's different.

Not a blocker, because I think it can be amended after the fact: there's some inconsistency with how the ranges are named in fields vs explained in docs, as noted in comments.

charts/region/crds/region.unikorn-cloud.org_networks.yaml
```diff
@@ -67,6 +67,32 @@ spec:
...
                 type: string
+              reservations:
+                description: |-
+                  Reservations allow parts of the address space to be reserved
+                  mainly for file storage but also VIP pools.
+                properties:
+                  nonStoragePrefixLength:
+                    description: |-
+                      NonStoragePrefixLength defines how much of the prefix to
+                      reserve for VIP pools, the remainder becomes the range
```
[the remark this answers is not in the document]
spjmurray: Well some of our users certainly need to be able to preserve .2-.15 as VIP space, so we're essentially advocating for a reservation, with some explicit user only stuff, and other services can consume the left overs as seen fit.

Which leads to a problem with if you wish to expand a use, then some ranges will most likely need to be moved potentially to allow the embiggened one to fit, and those services need to be notified of said changes and somehow react accordingly.  This is getting pretty complicated pretty quick...

spjmurray:
After further discussion, we have decided to keep this PR compact.  As a result there will be no implicit storage range defined here.  We will need product input and can address this in a follow up.

pkg/apis/unikorn/v1alpha1/types.go
```diff
@@ -446,12 +447,44 @@ type NetworkSpec struct {
 	Provider Provider `json:"provider,omitempty"`
 	// Prefix is the IPv4 address prefix.
 	Prefix *unikornv1core.IPv4Prefix `json:"prefix"`
+	// Reservations reserve a prefix at the start of the network for
+	// infrastructure use such as file storage and internal platform use
+	// as directed by the infrastructure provider.
+	// For example, on a /24 network a reservation prefix length of 25
+	// reserves 192.168.0.0/25, leaving 192.168.0.128-192.168.0.254 for DHCP.
+	// On create, omitting reservations means no reservations for newly-created
```
nsricardor: This comment contradicts the behavior: it says omitted reservations mean “no reservations for newly-created networks”, but the code applies the default /25 plus /28 carve-out.

nsricardor (dismissed):
Thanks, for updating this PR @spjmurray 

I’ve confirmed that the new behavior satisfies the storage reservation requirements:
* The default reservation keeps the current behavior, so this should not introduce a breaking change.
* Users can define their own storage reservation range during network creation.

PR #455, opened 2026-03-17
Generate secure passwords for OpenStack

yankcrime:
Instead of relying on a UUID, generate a (more) cryptographically secure password and crucially one that is shorter in length, thus saving a ton of noise in the Keystone logs regarding passwords being too long.

### DIFF
--- a/pkg/providers/internal/openstack/provider.go
+++ b/pkg/providers/internal/openstack/provider.go
@@ -19,6 +19,8 @@ package openstack
 
 import (
 	"context"
+	"crypto/rand"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"math/big"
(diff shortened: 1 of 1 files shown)


spjmurray (commented):
Magic numbers == bad, Claude can surely add a command line option, or add it to the Region CRD.

Other thing to be aware of is performance, but that's not a concern now.

yankcrime:
Closing this as I don't believe this to be the root cause of noise in Keystone logs.  The UUIDs we generate for passwords are 32 characters in length, whereas 72 characters should be the limit before any warnings are emitted (default `max_password_length` in Keystone is 4096 characters, but the default password hasher is bcrypt which truncates to 72 bytes).

PR #460, opened 2026-04-09
Add SSH certificate authority support for Region servers

spjmurray:
## Summary

This PR adds SSH certificate authority support for Region servers and carries the supporting API, CRD, RBAC, and provisioning changes needed for CA-backed launches to work end to end.

### Add the SSH CA resource surface
- add `SSHCertificateAuthority` to the Region resource surface and generated CRDs
- add API-side RBAC for listing, creating, and deleting SSH certificate authorities
- expose the CA ID on the Region server CRD so Kubernetes does not prune it during persistence

### Extend the v2 server API for SSH certificate authorities
- allow server create requests to reference an SSH certificate authority
- expose the configured CA ID in server status
- enforce that referenced CAs live in the same organization and project as the server
- reject malformed allowed source CIDRs with a clean `422` response instead of relying on schema validation
- use the correct `Update` RBAC permission for `UpdateV2`
- add handler tests covering cross-project CA rejection, invalid CIDR rejection, and preservation of immutable CA state on update

### Bootstrap SSH CA trust during provisioning
- add CA-aware cloud-init generation so CA-backed launches configure guest trust instead of relying on Nova key injection
- suppress legacy Nova `key_name` injection when a CA is configured
- add the controller RBAC needed to watch and update SSH CA references during reconciliation
- preserve server status mutations when cloud-init injection requires a create-time copy
- fix cloud-init serialization so generated config uses valid `write_files` and `runcmd` keys
- share cloud-init user-data recognition between API validation and provisioning logic
- add regression coverage around provider copy semantics and SSH CA cloud-init rendering

## Notes

- The API validates that a referenced CA exists when the server is created, but provisioning resolves it again later. If the CA disappears in that gap, reconciliation should fail the server into an error state rather than silently proceeding with the wrong trust model.


squaremo:
> Tests don't follow the required BDD structure from CLAUDE.md

This is a problem in CLAUDE.md, not here.

squaremo:
> SSHCertificateAuthorityID is in ServerSpec but not exposed in ServerStatus

I think the machine is confused here; the `SSH...ID` is returned in the status part of the response, but isn't in the status part of the CRD -- it's copied from the CRD _spec_ into the response _status_. No problem with the K8s API here.

charts/region/crds/region.unikorn-cloud.org_servers.yaml
```diff
@@ -103,6 +103,7 @@ spec:
                 enum:
                 - openstack
                 - kubernetes
+                - simulated
```
squaremo, line 106: This looks like a hold-over from a previous change rather than part of this change.
When you notice this kind of thing, squirrel it away into another branch.

pkg/handler/server/client_v2.go
```diff
@@ -216,17 +219,20 @@ func generateSecurityGroups(in *openapi.ServerV2Networking) []regionv1.ServerSec
...
+		return nil, nil
 	}
 
 	prefixes := *in.AllowedSourceAddresses
 
 	out := make([]regionv1.ServerNetworkAddressPair, len(prefixes))
 
 	for i := range prefixes {
-		_, prefix, _ := net.ParseCIDR(prefixes[i])
+		_, prefix, err := net.ParseCIDR(prefixes[i])
```
squaremo, line 232: Another unrelated change.

pkg/handler/server/client_v2.go
```diff
@@ -462,7 +516,7 @@ func (c *ClientV2) UpdateV2(ctx context.Context, serverID string, request *opena
 		return nil, err
 	}
 
-	if err := rbac.AllowProjectScope(ctx, "region:servers", identityapi.Delete, current.Labels[coreconstants.OrganizationLabel], current.Labels[coreconstants.ProjectLabel]); err != nil {
+	if err := rbac.AllowProjectScope(ctx, "region:servers", identityapi.Update, current.Labels[coreconstants.OrganizationLabel], current.Labels[coreconstants.ProjectLabel]); err != nil {
```
squaremo, line 519: Another unrelated change. (I thought I'd fixed this, maybe it was another copypasta)

pkg/handler/server/client_v2.go
```diff
@@ -476,7 +530,9 @@ func (c *ClientV2) UpdateV2(ctx context.Context, serverID string, request *opena
 		return nil, err
 	}
 
-	required, err := c.generateV2(ctx, current.Labels[coreconstants.OrganizationLabel], current.Labels[coreconstants.ProjectLabel], request, network)
+	// User data is only consumed during initial server bootstrap. Updates preserve it for
+	// completeness and future rebuild support, but they do not re-run cloud-init validation.
```
squaremo, line 534: To be clear though, you can't update it once you've created the machine, right?
(Then you'd have a record of the user data that didn't match what the machine was booted with, which would just be confusing)

squaremo (approved):
How this new API surface works seems fine :+1: And there are reasonable-looking tests for most of the moving parts (the difficult one is checking that the userdata actually does what it's supposed to, of course). With that and the machine's review, I thing this is good to go.
Please see the notes below for thoughts on making this easier to review in depth.

---
This is too big for me to review in depth without it being rather time-consuming (time being a scarce resource), so I didn't do that.

- there are some unrelated changes in here, which could have been done as "ship" PRs (i.e., just merged)
- userdata validation could go in its own change, as well -- it adds (theoretical) value on its own, and that makes this change smaller.

Pulling those out wouldn't drastically alter the size of this change -- only by a few hundred lines. Changes that add to the model will always be large overall. They could be done, reviewed, tested and merged from the back forwards, i.e., provider first, then provisioner, then client and API .. and that's what we should really aim for, since each one of those changes would be much easier to review (and reread, and understand individually) in depth.

PR #461, opened 2026-04-09
feat(monitor): emit structured log on server state transition

peterj-ns:
Emits a structured log line in the region monitor whenever a server's lifecycle phase changes, enabling Loki queries for state change history per instance, org, and region.

## Changes

- `pkg/monitor/health/server/check.go` — trigger on `Status.Phase` change after `Status().Patch()` in `checkServer()`; `time_since_creation_ms` is measured from `CreationTimestamp` (there is no per-phase transition timestamp in the schema)
- `pkg/monitor/health/server/check_test.go` — unit tests covering: phase change fires the log with all required fields, unchanged phase produces no log, condition-only change produces no log

## Validation

1. Deploy the updated monitor binary
2. Stop a running server via the API
3. Query Loki: `{app="unikorn-region-monitor"} |= "instance state transition" | json`
4. Confirm fields present: `instance_id`, `org_id`, `region_id`, `from_state`, `to_state`, `time_since_creation_ms`

## Notes

`time_since_creation_ms` measures time since server creation, not time in the previous phase — there is no `PhaseTransitionTime` field on `ServerStatus` to anchor it more precisely.

Closes INST-725

### DIFF
--- a/pkg/monitor/health/server/check.go
+++ b/pkg/monitor/health/server/check.go
@@ -21,13 +21,16 @@ import (
 	"context"
 	goerrors "errors"
 	"fmt"
+	"time"
 
+	coreconstants "github.com/unikorn-cloud/core/pkg/constants"
 	"github.com/unikorn-cloud/core/pkg/errors"
 	unikornv1 "github.com/unikorn-cloud/region/pkg/apis/unikorn/v1alpha1"
 	"github.com/unikorn-cloud/region/pkg/constants"
 	"github.com/unikorn-cloud/region/pkg/providers"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 // Checker for server health.
@@ -84,6 +87,17 @@ func (c *Checker) checkServer(ctx context.Context, server *unikornv1.Server) err
 
 	if err := c.client.Status().Patch(ctx, updated, client.MergeFromWithOptions(server, client.MergeFromWithOptimisticLock{})); err != nil {
 		return err
+	}
+
+	if server.Status.Phase != updated.Status.Phase {
+		log.FromContext(ctx).Info("instance state transition",
+			"instance_id", server.Name,
+			"org_id", server.Labels[coreconstants.OrganizationLabel],
+			"region_id", server.Labels[constants.RegionLabel],
+			"from_state", server.Status.Phase,
+			"to_state", updated.Status.Phase,
+			"time_since_creation_ms", time.Since(server.CreationTimestamp.Time).Milliseconds(),
+		)
 	}
 
 	return nil
--- a/pkg/monitor/health/server/check_test.go
+++ b/pkg/monitor/health/server/check_test.go
@@ -1,0 +1,224 @@
+/*
+Copyright 2026 Nscale.
+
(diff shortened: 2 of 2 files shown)


pkg/monitor/health/server/check.go
```diff
@@ -86,6 +90,23 @@ func (c *Checker) checkServer(ctx context.Context, server *unikornv1.Server) err
 		return err
 	}
 
+	if server.Status.Phase != updated.Status.Phase {
+		durationSource := server.CreationTimestamp.Time
+
+		if condition, err := server.StatusConditionRead(unikornv1core.ConditionHealthy); err == nil {
```
squaremo: The phase is not the same thing as the health condition, and I'm not convinced they are changed _at the same time_. So this is going to give you an inaccurate picture. If you can show they get changed at the same time (doesn't look like it to me, at a scan), fine. But otherwise, you might need to check that the condition value is different, rather than the phase.

squaremo (changes requested):
Reasonable idea; the phase and the health condition are not the same thing and may not get changed at the same time though. Either prove they do get changed at the same time, or report when the condition changes instead.

pkg/monitor/health/server/check.go
```diff
@@ -86,6 +90,23 @@ func (c *Checker) checkServer(ctx context.Context, server *unikornv1.Server) err
...
+			durationSource = condition.LastTransitionTime.Time
+		}
```
squaremo: I thought you were going to remove this -- since the condition change does not correspond with the phase change -- in favour of consistently logging the time since creation.

pkg/monitor/health/server/check.go
```diff
@@ -86,6 +89,17 @@ func (c *Checker) checkServer(ctx context.Context, server *unikornv1.Server) err
...
 	}
 
+	if server.Status.Phase != updated.Status.Phase {
+		log.FromContext(ctx).Info("instance state transition",
+			"instance_id", server.Name,
+			"org_id", server.Labels[coreconstants.OrganizationLabel],
+			"region_id", server.Labels[constants.RegionLabel],
+			"from_state", server.Status.Phase,
+			"to_state", updated.Status.Phase,
+			"duration_ms", time.Since(server.CreationTimestamp.Time).Milliseconds(),
```
squaremo: Is there a reason this is called `duration_ms` rather than, say, `time_since_creation_ms`? It _is_ a duration, but it doesn't measure the duration people are going to presume it does.

pkg/monitor/health/server/check.go
peterj-ns: Updated - sorry forgot to push changes.

pkg/monitor/health/server/check.go
peterj-ns: Updated to `time_since_creation_ms` to reduce potential confusion.

peterj-ns:
> Reasonable idea; the phase and the health condition are not the same thing and may not get changed at the same time though. Either prove they do get changed at the same time, or report when the condition changes instead.

Yep, simplified approach to just logging with creation timestamp.

squaremo (approved):
Looks reasonable. We'll get a few duplicate entries, e.g., when the process restarts, but I don't think that will be much of a hindrance, and I'd rather have that than more complexity :-)

PR #470, opened 2026-04-16
Add OTel server state and provisioning duration metrics

peterj-ns:
## Summary

Adds two OTel metrics to the region server monitor:

- `unikorn_region_server_state` (`Int64ObservableGauge`) — server counts by `{state, region_id, region_name, flavor_id, flavor_name}`, rebuilt from the full server list on each poll cycle
- `unikorn_region_server_provision_duration_seconds` (`Float64Histogram`) — Pending duration observed on Pending → Running, with `{region_id, region_name, flavor_id, flavor_name}` labels and buckets `[5, 15, 30, 60, 120, 300, 600, 900, 1800, 3600]`

## Implementation notes

A `regions.unikorn-cloud.org/pending-entry-time` annotation is written when a server enters Pending and removed when it leaves. This bounds provisioning duration measurement to the current Pending period only — accurate across stop/start cycles and pod restarts. A `patch` rule on the `servers` resource is added to the monitor ClusterRole.

State counts use the post-transition phase (propagated back after each successful status patch) so the gauge reflects updated state immediately within each poll cycle. Because the observable gauge callback reports only what is in the current snapshot, stale entries do not carry forward across poll cycles. Servers skipped due to region resolution or provider errors are absent from the gauge for that cycle rather than reported as unknown.

Each metric emits both the stable identifier (`region_id`, `flavor_id`) and the human-readable display name (`region_name`, `flavor_name`) per the platform metrics label guidance. The identifier is authoritative for alerting; the name is for dashboard legends. If name resolution fails, the name label is emitted as an empty string. Region and flavor names are resolved once per unique region per poll cycle by caching the provider `Region()` and `Flavors()` results. Resolution failures are cached to avoid hammering a failing provider on every server in the region.

The histogram buckets extend to 3600 s (1 h) to cover GPU provisioning workloads that can exceed 15 minutes.

Full metric export requires nscaledev/uni-core#288 to be deployed; until then the meter is a no-op and metrics are recorded in-process only.

### Automated

- `pkg/monitor/health/server/metrics_test.go` — gauge emits correct observations and stale entries don't persist across replacements; histogram records correct attributes
- `pkg/monitor/health/server/check_test.go` — phase and health condition transition logging; annotation stopwatch (stamp on entry, no overwrite, removal on exit, fresh stamp on re-entry); histogram recorded through `Check` with correct `region_id`/`region_name`/`flavor_id`/`flavor_name` attributes; name resolves to empty string on provider error or miss; integration test wiring `Check` → gauge asserts lowercase state labels and correct counts

### Manual

1. Query `unikorn_region_server_state` in Grafana — counts should match actual server states, labelled by region and flavor ID and display name
2. Provision a server and confirm `unikorn_region_server_provision_duration_seconds` has an observation once it reaches Running
3. Stop and restart a server — confirm the histogram records the re-provisioning duration, not total server age


squaremo:
> * `unikorn_region_server_state` (`Int64ObservableGauge`) — server counts by `{state, region}`, rebuilt from the full server list on each poll cycle
> unikorn_region_server_provision_duration_seconds (Float64Histogram) — Pending duration observed on Pending → Running, with {region} label and buckets [30, 60, 120, 300, 600, 900]

I think you'd want to also label these by flavour.

For the first one I can imagine it being useful to be able to break the count down that way.

For the second, VMs will provision much faster than bare metal. This would skew the distribution (VM observations will all end up in the first bucket). You may also want a bucket or two at the bottom end, so that you can cover what's going to be a pretty strongly bimodal distribution (unless you want to have two metrics, one for bare metal and one for VMs).

pkg/monitor/health/server/check.go
```diff
@@ -41,17 +44,127 @@ type Checker struct {
...
+
+	if updated.Status.Phase != unikornv1.InstanceLifecyclePhasePending {
+		if !hasAnnot {
+			return nil
+		}
+
+		patch := updated.DeepCopy()
+		delete(patch.Annotations, constants.ServerPendingEntryTimeAnnotation)
+
+		return c.client.Patch(ctx, patch, client.MergeFrom(updated))
```
spjmurray: Never call patch without optimistic locking, it breaks the model that we know about things going wrong, rather than silently overwritten.

cmd/unikorn-region-monitor/main.go
```diff
@@ -70,12 +70,20 @@ func main() {
 		cancel()
 	}()
 
+	if err := monitorOptions.CoreOptions.SetupOpenTelemetry(ctx); err != nil {
+		logger.Error(err, "failed to setup OpenTelemetry")
+
+		return
```
squaremo: Should you os.Exit(1) here, so it's recorded as a crash? This may have a bearing on what the pod controller does.

pkg/monitor/health/server/check.go
peterj-ns: updated.

pkg/monitor/health/server/metrics.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// collectStateCounts is the OTel observable callback for unikorn_region_server_state.
+// It is called by the SDK on each metric collection cycle.
+func (m *Metrics) collectStateCounts(ctx context.Context, o metric.Observer) error {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	for k, count := range m.stateCounts {
+		o.ObserveInt64(m.stateGauge, count,
```
squaremo, line 95: Are counts that are not observed in the callback assumed to be zero?

squaremo: If not, you may need to explicitly set previous counts to zero then drop them. You'll be able to tell if it's a problem if one Server changes state but you have a count of `1` for both states in the metric.

pkg/monitor/health/server/check.go
```diff
@@ -41,17 +44,127 @@ type Checker struct {
...
+
+	// Server just entered Pending: stamp the current time.
+	patch := updated.DeepCopy()
+	if patch.Annotations == nil {
+		patch.Annotations = make(map[string]string)
+	}
+
+	patch.Annotations[constants.ServerPendingEntryTimeAnnotation] = time.Now().UTC().Truncate(time.Second).Format(time.RFC3339)
+
+	return c.client.Patch(ctx, patch, client.MergeFrom(updated))
```
squaremo: Here too (optimistic lock)

pkg/monitor/health/server/check.go
```diff
@@ -41,17 +44,127 @@ type Checker struct {
...
+		return nil
+	}
+
+	// Server just entered Pending: stamp the current time.
+	patch := updated.DeepCopy()
+	if patch.Annotations == nil {
+		patch.Annotations = make(map[string]string)
+	}
+
+	patch.Annotations[constants.ServerPendingEntryTimeAnnotation] = time.Now().UTC().Truncate(time.Second).Format(time.RFC3339)
```
squaremo: I don't think you need to truncate here, since the format is going to drop fractions of seconds anyway.

pkg/monitor/health/server/check.go
```diff
@@ -89,17 +202,18 @@ func (c *Checker) checkServer(ctx context.Context, server *unikornv1.Server) err
...
+	c.logStateTransition(ctx, server, updated)
+
+	if err := c.managePendingAnnotation(ctx, updated); err != nil {
+		// Annotation patch failures are non-fatal: the next poll will re-attempt.
+		// Aborting the loop here would block all subsequent servers for a transient error.
+		log.FromContext(ctx).Error(err, "failed to patch pending-entry annotation", "server", server.Name)
 	}
 
+	// Propagate the updated phase back so updateStateCounts reflects post-transition state.
+	server.Status.Phase = updated.Status.Phase
```
squaremo: I do not like this mutation, since the value comes from the cache and could upset other references. Perhaps you could return updated|server and build the state count from that list.

squaremo (changes requested):
Things to consider:
- os.Exit(1) when there's an error
- what happens to state counts for which you don't have an observation, are they reset to zero?
- avoiding mutating the cached Server value

Things to change:
- needs optimistic lock in the other call to client.Patch

pkg/monitor/health/server/check.go
```diff
@@ -41,66 +44,232 @@ type Checker struct {
...
+
+		seen[regionID] = true
+
+		provider, err := c.providers.LookupCloud(regionID)
+		if err != nil {
+			log.FromContext(ctx).Error(err, "failed to look up provider for flavour names", "region", regionID)
+			continue
+		}
+
+		flavours, err := provider.Flavors(ctx)
```
squaremo: I would not do this inside the loop.

pkg/monitor/health/server/metrics.go
peterj-ns: The callback is a complete snapshot so the SDK only reports what's observed each cycle and doesn't carry forward previous values. We can see this in `TestSetStateCountsReplacesMap` where swapping the map to a single new entry results in exactly one data point on the next collection, rather than three.

pkg/monitor/health/server/check.go
peterj-ns: Updated.

pkg/monitor/health/server/check.go
peterj-ns: Updated so now we resolve both names inline in `checkServer` where we already have the provider in scope.

cmd/unikorn-region-monitor/main.go
peterj-ns: Updated

pkg/monitor/health/server/check.go
peterj-ns: Done — so `checkServer` now returns `updated` and results are collected into an effective []checkedServer slice; then `updateStateCounts` builds counts from that directly with no mutation.

peterj-ns:
> Things to consider:
> 
> * os.Exit(1) when there's an error
> * what happens to state counts for which you don't have an observation, are they reset to zero?
> * avoiding mutating the cached Server value
> 
> Things to change:
> 
> * needs optimistic lock in the other call to client.Patch

All four points addressed — see individual comment replies.

squaremo:
@peterj-ns, would you
- include the ID of flavours and regions as labels (see the draft guidance at https://github.com/nscaledev/uni-specifications/pull/39)
- rebase on main

Thanks!

peterj-ns:
> @peterj-ns, would you
> 
> * include the ID of flavours and regions as labels (see the draft guidance at [Guidance on metrics labels uni-specifications#39](https://github.com/nscaledev/uni-specifications/pull/39))
> * rebase on main
> 
> Thanks!

Done — both addressed!                                                                                                                                                         
                                                                                                                                                                                  
The metrics now emit both the stable identifier and display name as separate labels per the spec guidance (_id and _name suffixes). The gauge and histogram both carry `region_id`, `region_name`, `flavor_id`, and `flavor_name`.

squaremo (approved):
I'm happy the things I brought up have been addressed. (I still think baremetal and VMs should have different histograms because they'll have very different distributions ... but I guess we'll see :-))
Thanks Peter! 🍍

PR #471, opened 2026-04-17
Regenerate CRDs

spjmurray:
The robots seem to have lost the CRD update that enables SSH CA functionality.  Coupled with the fact that the CI jobs didn't spot it either.  This essentially adds in the SSH CA field to server CRD.


squaremo:
> The CRD should have been regenerated then but wasn't

I think the purity test does not run the target that updates CRDs in the chart.

PR #474, opened 2026-04-20
bump

spjmurray:
- **openstack: add temporary image cache write-through bridge**
- **Version v1.16.0-rc3.2**

### DIFF
--- a/charts/region/Chart.yaml
+++ b/charts/region/Chart.yaml
@@ -4,8 +4,8 @@ description: A Helm chart for deploying Unikorn's Region Controller
 
 type: application
 
-version: 1.16.0-rc3.1
-appVersion: 1.16.0-rc3.1
+version: 1.16.0-rc3.2
+appVersion: 1.16.0-rc3.2
 
 icon: https://raw.githubusercontent.com/unikorn-cloud/assets/main/images/logos/dark-on-light/icon.png
 
--- a/pkg/providers/internal/openstack/export_test.go
+++ b/pkg/providers/internal/openstack/export_test.go
@@ -39,6 +39,19 @@ func NewImageQuery(listFunc func() (*cache.ListSnapshot[types.Image], error)) ty
 	return &imageQuery{listFunc: listFunc}
 }
 
+type ImageCacheBase = imageCacheBase
+
+type ImageCacheWrapper = imageCacheWrapper
+
+//nolint:gochecknoglobals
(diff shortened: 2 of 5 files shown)


spjmurray:
Follow-up on the latest Claude review, focusing only on the points that are still relevant after the pending_delete fix.

1. Test structure
In this repository Ginkgo is used for integration/API tests, not for ordinary unit tests. The wrapper tests are intentionally small, isolated unit tests for a temporary adapter, so plain testing is deliberate here. We are not going to rewrite disposable unit tests into BDD-style Ginkgo just to satisfy that comment.

2. unsafe / reflect in newCacheEpoch
This is ugly, but with the current cache API there is no clean external alternative if we want to preserve epoch semantics correctly.

The wrapper must advance the observable epoch when local InsertIfAbsent / Update changes the effective image view before the underlying refresh-ahead cache epoch changes. From outside the cache package we currently have:
- an exported cache.Epoch type
- no exported fields
- no constructor
- no helper to derive a new epoch

So from outside the package the choices are basically:
- manufacture an epoch via unsafe / reflect
- incorrectly reuse the base epoch and break the epoch contract
- introduce a parallel epoch type and widen APIs further
- move this functionality into the cache package itself

For a temporary adapter, the isolated unsafe helper was the least bad option. This should disappear when the mutation support is folded back into the shared cache.

3. Coverage
We are not treating Codecov patch coverage as a blocker for this change. The wrapper is intentionally temporary and isolated. The main verification bar here was targeted unit coverage for the wrapper semantics plus lint passing.

4. Get() using List()
This is an intentional tradeoff forced by the current cache API, not an oversight.

The wrapper lifecycle is keyed off "first observed underlying cache epoch change", because that is what clears the temporary bridge state. base.Get(id) does not give enough information to do that correctly. The wrapper needs a coherent snapshot epoch from the underlying cache and, with the current API, List() is the only operation that gives that in a way we can use.

So the realistic choices today were:
- back Get() with List() and keep the agreed semantics
- weaken the wrapper semantics
- change the shared cache API first

Given the release pressure, we chose the first option deliberately. Like the epoch hack, this is a temporary consequence of the missing native cache mutation API and is expected to go away when this logic is folded into the shared cache.

5. Removal plan
Agreed in principle that this should not live forever. The intent remains to move InsertIfAbsent / Update semantics into the shared cache and then delete this wrapper.

PR #475, opened 2026-04-20
Add v2 load balancer API and integration coverage

mkjpryor:
## Summary
Adds the v2 load balancer API surface, including the OpenAPI spec, generated router/client/types, CRD registration, and handler wiring for list/create/get/update/delete operations. It implements the load balancer v2 client logic for filtering, conversion, validation, and lifecycle management, and updates RBAC to expose the new endpoint. It also extends the typed integration test helpers with load balancer fixtures/endpoints and adds an integration suite that exercises create/list/get/update/delete behavior, alongside the related contract fixture updates.


charts/region/crds/region.unikorn-cloud.org_loadbalancers.yaml
squaremo, line 1: Looks reasonable, and Nick thinks it looks reasonable too, so...

test/contracts/consumer/identity/identity_test.go
squaremo, line 1: > Also picks up a pact-matcher refactor in the identity consumer test
that was required by the regenerated identity types.

What happened here? This doesn't seem like it's at all connected to this change (either this commit, or the whole PR).

pkg/handler/loadbalancer/client_v2_test.go
```diff
@@ -0,0 +1,901 @@
...
+	lbAllocationID   = "allocation-1"
+)
+
+func newLBFakeClientBuilder(t *testing.T, objects ...runtime.Object) *fake.ClientBuilder {
+	t.Helper()
+
+	scheme := runtime.NewScheme()
+	require.NoError(t, regionv1.AddToScheme(scheme))
+
+	restMapper := apimeta.NewDefaultRESTMapper([]schema.GroupVersion{regionv1.SchemeGroupVersion})
```
squaremo, line 79: This stuff isn't usually necessary (e.g., see https://github.com/nscaledev/uni-region/blob/8a639afc8b9a64be288db34a2161b3882d42ebd8/pkg/handler/network/client_v2_test.go#L89), how come this is a special case?

pkg/handler/handler_v2.go
```diff
@@ -174,6 +175,73 @@ func (h *Handler) DeleteApiV2SecuritygroupsSecurityGroupID(w http.ResponseWriter
 	w.WriteHeader(http.StatusAccepted)
 }
 
+func (h *Handler) loadBalancerClient() *loadbalancer.Client {
```
squaremo, line 178: I'd prefer to see this pulled out into at least their own file (more a tidiness concern than anything), or their own handler struct (this helps with isolating tests); as an example, https://github.com/nscaledev/uni-region/blob/8a639afc8b9a64be288db34a2161b3882d42ebd8/pkg/handler/handler_v2_image.go.

squaremo (dismissed):
I'm happy for this to get merged, since the API looks right, there are tests at different levels with fair coverage, and anyway there's opportunity to correct things before it's made available.

It looks like the API tests depend on some changes to RBAC roles that haven't been deployed; merging this might make that a problem for every subsequent run of those tests, so that should be fixed first, or at least we should be confident it's going to be fixed soon.

There's an apparently unrelated change in here which I'd rather not have mixed in (it makes diagnosing where problems happen, and rolling changes back, harder). I'd like to see it obviated by a PR that's merged first, or at least in its own commit with an explanation. But I'm not making that a precondition for merging, up to you.

peterj-ns:
The integration test failure is due to the missing RBAC role `region:loadbalancers:v2` not being there which is what's causing the 403. so merging [PR 453](https://github.com/nscaledev/uni-identity/pull/453) is necessary but you'll also need to bump the identity dependency in go.mod on your branch to pick it up, otherwise the integration test installs the old pinned version and won't see the change.

pkg/handler/handler_v2.go
mkjpryor: SSH CA client and storage client follow the same pattern - best to fix them all together separately?

squaremo: Yeh I did that and it got ignored. So, we go incremental.

pkg/handler/loadbalancer/client_v2_test.go
```diff
@@ -0,0 +1,904 @@
...
+func newLBFakeClientBuilder(t *testing.T, objects ...runtime.Object) *fake.ClientBuilder {
+	t.Helper()
+
+	scheme := runtime.NewScheme()
+	require.NoError(t, regionv1.AddToScheme(scheme))
+
+	// These tests exercise identity allocation creation/update, which calls
+	// manager.GenerateResourceReference and needs RESTMapping for custom
+	// resources. The fake client defaults to an empty REST mapper.
+	restMapper := apimeta.NewDefaultRESTMapper([]schema.GroupVersion{regionv1.SchemeGroupVersion})
```
squaremo, line 79:
```suggestion
    // a restMapper with entries for referenced types is assumed by manager.GenerateResourceReference
	restMapper := apimeta.NewDefaultRESTMapper([]schema.GroupVersion{regionv1.SchemeGroupVersion})
```

squaremo (dismissed):
Rubber stamp ✅

pkg/apis/unikorn/v1alpha1/types.go
```diff
@@ -638,6 +638,128 @@ type SecurityGroupRule struct {
...
+}
+
+type LoadBalancerSpec struct {
+	// Pause, if true, will inhibit reconciliation.
+	Pause bool `json:"pause,omitempty"`
+	// Tags are an abitrary list of key/value pairs that a client
+	// may populate to store metadata for the resource.
+	Tags unikornv1core.TagList `json:"tags,omitempty"`
+	// RequestedVIPAddress is the optional requested virtual IP address.
+	RequestedVIPAddress *unikornv1core.IPv4Address `json:"requestedVIPAddress,omitempty"`
```
spjmurray, line 672: Is this VIP limited to private IPs from the underlying network?  Is it from the DHCP pool only, or can it be from the static "VIP" allocation?  Reason I'm asking is it's a documentation gap I cannot be bothered to manually look up 😄

That obviously extends to the public facing documentation in the openapi specs.

This then leads me to the point that there should be some error checking expected in the handler for both this VIP, and any pool members so we can fail fast to the user with some useful feedback.

spjmurray (commented):
Just one comment to address, the rest is boilerplate so I'm placing trust in the machine!

pkg/apis/unikorn/v1alpha1/types.go
mkjpryor: Agree that this IP, if given, should be validated as being in the owning network's prefix. Pool members are explicitly not subject to this validation as the only constraint on those is that they are reachable from the LB.

pkg/openapi/server.spec.yaml
```diff
@@ -2596,6 +2724,200 @@ components:
...
+          items:
+            $ref: '#/components/schemas/loadBalancerListenerV2'
+    loadBalancerV2CreateSpec:
+      description: A load balancer creation specification.
+      type: object
+      allOf:
+      - $ref: '#/components/schemas/loadBalancerV2Spec'
+      - type: object
+        required:
+        - organizationId
```
spjmurray: organizationId/projectId/regionId are all redundant here, and are not part of other V2 APIs e.g. security groups.  All of these parameters can be inferred from the network, and hence it's unnecessary user input.

pkg/handler/loadbalancer/client_v2.go
```diff
@@ -0,0 +1,930 @@
...
+	}
+
+	if err := controllerutil.SetOwnerReference(network, out, c.Client.Scheme(), controllerutil.WithBlockOwnerDeletion(true)); err != nil {
+		return nil, fmt.Errorf("%w: unable to set resource owner", err)
+	}
+
+	return out, nil
+}
+
+func validateCreateRequestMatchesNetwork(request *openapi.LoadBalancerV2Create, network *regionv1.Network) error {
```
spjmurray: This check is redundant.  When the Create handler calls networks.GetV2Raw that will perform the necessary RBAC checks, and more pertinently function correctly as that user directly or as that user indirectly via impersonation.  Also taking into account RBAC permissions, not explicitly granting them as we have here.

spjmurray (commented):
RIght, high level philosophy is a pass.  Low level technical stuff, I have a couple comments...

pkg/handler/loadbalancer/client_v2.go
```diff
@@ -0,0 +1,910 @@
...
+	out := &openapi.LoadBalancerV2Update{}
+
+	if err := json.Unmarshal(t, out); err != nil {
+		return nil, fmt.Errorf("%w: failed to unmarshal request", err)
+	}
+
+	return out, nil
+}
+
+func validatePort(port int, field string) error {
```
spjmurray, line 286: I like how defensive it is, it's already been checked by JSON schema...

pkg/handler/loadbalancer/client_v2.go
```diff
@@ -0,0 +1,910 @@
...
+	ip, err := parseIPv4Address(*in, "vipAddress")
+	if err != nil {
+		return unikornv1core.IPv4Address{}, false, err
+	}
+
+	return unikornv1core.IPv4Address{IP: ip}, true, nil
+}
+
+func validateRequestedVIPAddressMatchesNetwork(vipAddress *unikornv1core.IPv4Address, network *regionv1.Network) error {
+	if vipAddress == nil {
```
spjmurray, line 435: Flat code!  Early exits ❤️

spjmurray (approved):
I think we're good to go... you have manually tested creation, update and delete via the APIs right as a final sanity check right, ensured things are correctly reflected in the CRs?

spjmurray:
> manually tested creation, update and delete via the APIs 

This is where Unikorn UI earns its keep, just FYI...

mkjpryor:
Local testing has been done, `uni-ui` change in development.

PR #478, opened 2026-04-22
Use GOBIN-installed tools in Makefile targets

mkjpryor:
This updates Makefile targets to invoke installed tools from `$(GOBIN)` instead of relying on whatever happens to be on the shell PATH. It also prepends `$(GOBIN)` to the PATH used by `go generate`, so generated code uses the freshly installed binaries consistently. The change covers OpenAPI generation and Ginkgo-based API test targets. Validation run locally: `make touch`, `make license`, `make validate`, `make lint`, `make generate`, and `make test-unit`.


squaremo (approved):
Approved on the basis of
- mechanical change
- tests pass in the isolated environment
- machines says it's OK

PR #483, opened 2026-04-23
Add load balancer controller scaffolding

mkjpryor:
## Summary

Introduces a skeleton controller for the v2 `LoadBalancer` CRD:

- New `unikorn-load-balancer-controller` binary, factory, and stub provisioner (`Provision` yields; no provider wiring yet).
- Helm chart deployment with service account and RBAC (role, rolebinding, clusterrole, clusterrolebinding) mirroring the network controller.
- `Deprovision` releases the identity quota allocation created by the API handler saga, closing the quota leak for deleted load balancers.

## Test plan

- [ ] `make validate lint generate test-unit`
- [ ] `helm template charts/region ...` renders the new load-balancer-controller resources
- [ ] Existing `test/api/suites/loadbalancers_test.go` integration suite passes

### DIFF
--- a/charts/region/templates/load-balancer-controller/clusterrole.yaml
+++ b/charts/region/templates/load-balancer-controller/clusterrole.yaml
@@ -1,0 +1,47 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ .Release.Name }}-load-balancer-controller
+  labels:
+    {{- include "unikorn.labels" . | nindent 4 }}
+rules:
+# Orchestrate Unikorn resources (my job).
+- apiGroups:
+  - region.unikorn-cloud.org
+  resources:
+  - identities
+  - openstackidentities
+  verbs:
+  - list
+  - watch
+- apiGroups:
+  - region.unikorn-cloud.org
+  resources:
+  - loadbalancers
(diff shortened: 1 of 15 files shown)


charts/region/templates/load-balancer-controller/clusterrole.yaml
```diff
@@ -0,0 +1,47 @@
...
+metadata:
+  name: {{ .Release.Name }}-load-balancer-controller
+  labels:
+    {{- include "unikorn.labels" . | nindent 4 }}
+rules:
+# Orchestrate Unikorn resources (my job).
+- apiGroups:
+  - region.unikorn-cloud.org
+  resources:
+  - identities
```
spjmurray, line 12: I swear we are missing networks here so you can translate from unikorn -> openstack

spjmurray: Talk through it with the AI, "given we need to call this openstack API find out the parameters we will need and where we need to source them from given our CRD layout".  That kind of jazz.

mkjpryor: the controller doesn't wait for the network yet... that is coming in a future ticket.

spjmurray (approved):

PR #485, opened 2026-04-23
Add provider-layer load balancer interface

mkjpryor:
## Summary
Add load balancer create/delete methods to the provider contract and regenerate the provider mocks.
Implement deterministic simulated-provider load balancer VIP/public IP behavior with focused unit coverage for requested VIPs, derived VIPs, public IP allocation, idempotent delete, and missing-network failures.
Add temporary OpenStack load balancer stubs so the provider still satisfies the expanded interface until real support lands.

## Verification
make touch
make license
make validate
make lint
make generate
make test-unit

### DIFF
--- a/pkg/providers/internal/openstack/loadbalancer.go
+++ b/pkg/providers/internal/openstack/loadbalancer.go
@@ -1,0 +1,34 @@
+/*
+Copyright 2026 Nscale.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


mkjpryor:
Addressing Claude's review items 1 and 2:

1. I'm leaving `pkg/providers/internal/simulated/provider_test.go` unchanged. The new load balancer tests follow the existing `pkg/` unit-test style already used in that file and elsewhere under `pkg/`, rather than the Ginkgo style used in `test/api` integration suites.

2. I made the minimal code change for the simulated public IP prefix by documenting that `198.51.100.0/24` is RFC 5737 TEST-NET-2 and is intentionally used for deterministic, non-routable simulated public IP allocation.

Verification run:
- `go test ./pkg/providers/internal/simulated`
- `make touch`
- `make license`
- `make validate`
- `make lint`
- `make generate`
- `make test-unit`

spjmurray (approved):
Appears sane, it's prerelease, so let's see what happens.

PR #488, opened 2026-04-24
Pin the server host with infrastructureRef

squaremo:
To be able to place bare metal machines, we need to be able to tell the provider (OpenStack, really) which host has been chosen. This PR when merged will:

 - add a field infrastructureRef* to the Server type, and when provisioning, pass it to Nova so that it uses it to target a specific machine
 - accept the field on the server create endpoint

(*the term `infrastructureRef` is taken from OpenStack -- it's the name for an Ironic machine identifier)

The API tests are extended to check that a Server targeting a specific host can be created. Like the other "Server Management" will not run in GHA workflows until we can run them against a stable devstack. But they can be run locally if you provide a TEST_SERVER_INFRASTRUCTURE_REF alongside the other required entries, which `hack/openstack/configure-server` can do for you -- see the README.

EDIT: I split out `pinnedOnly` to #491.


squaremo:
~@claude review please~
(EDIT: I have not updated the triggers for automated review here yet)

squaremo:
@JohnGarbutt review request is FYI and optional :-)

pkg/apis/unikorn/v1alpha1/types.go
```diff
@@ -894,6 +894,11 @@ type ServerSpec struct {
 	SSHCertificateAuthorityID *string `json:"sshCertificateAuthorityID,omitempty"`
 	// UserData contains configuration information or scripts to use upon launch.
 	UserData []byte `json:"userData,omitempty"`
+	// InfrastructureRef pins the server to a specific physical host. When set,
+	// the provider bypasses its scheduler and provisions directly onto the
+	// identified host.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="infrastructureRef is immutable"
```
spjmurray: `self == oldSelf` on an optional field has a gap: Kubernetes only evaluates field-level transition rules when the field exists in the old object. If a server is created without `infrastructureRef`, a later update can set it — the rule is never evaluated for the absent→present transition.

Since updates are supported and a rebuild path is expected in future, this matters: silently setting `infrastructureRef` on an existing server would trigger a full teardown and reprovision onto a different host.

The fix is to move the validation to the `ServerSpec` level where `has()` can be used:

```go
// +kubebuilder:validation:XValidation:rule="!has(oldSelf.infrastructureRef) == !has(self.infrastructureRef) && (!has(self.infrastructureRef) || self.infrastructureRef == oldSelf.infrastructureRef)",message="infrastructureRef is immutable"
type ServerSpec struct {
```

This blocks changing the value, unsetting it, and setting it after creation.

squaremo: TIL about validation on optional fields ...

Via the API this exploit is prevented, since the update path doesn't set or unset that field. But that validation is a better guard nonetheless.

squaremo: Validation changed as suggested 🙇

spjmurray (approved):

PR #490, opened 2026-04-27
test: fail on teardown errors instead of logging warnings

peterj-ns:
Closes INST-848

Teardown failures are now test failures. Previously, cleanup errors were swallowed and logged as warnings, meaning deletion bugs (e.g. hanging deletes causing resource leaks) were invisible to the suite.

Resources deleted by a test spec nil out their ID so the `AfterAll` safety net skips the redundant delete on the happy path.


squaremo (approved):
Yep, failing to clean up is also a failure (albeit an annoying one :-D)

JakeNScale (approved):

PR #491, opened 2026-04-27
Guard for flavours that require a host ID

squaremo:
This adds a property to flavours (which can be set in the region object) which restricts that flavour to creating Servers with an `infrastructureRef` (a host ID particular to the provider).

This is to stop people creating bare metal machines without going through placement, and similar use cases. The converse is not true: this doesn't limit when you can use `infrastructureRef` otherwise. We _may_ want that restriction too, but for the purpose of placement, it's this way around.

The behaviour is backward-compatible, in that it behaves the same as now if the field is not set on a flavour.


squaremo:
> [pinnedOnly in a flavour spec] is to stop people creating bare metal machines without going through placement, and similar use cases. The converse is not true: this doesn't limit when you can use `infrastructureRef` otherwise. We _may_ want that restriction too, but for the purpose of placement, it's this way around.

@mkjpryor @spjmurray @JohnGarbutt Do you reckon we should do the converse as well: that is, _unless_ a flavour is marked as `pinnedOnly`, you _cannot_ supply a machine ID?

nsricardor (approved):

PR #492, opened 2026-04-28
Mention `manager` role in Openstack Setup Docs

spjmurray:
In order fo the provider to publicize images it needs to have a token scoped to a project (how Glance works).  It also needs the necessary permissions to be granted, which currently are injected via the `manager` role.  As such we need to update the documentation so that role gets bound to the provider user and project.

### DIFF
--- a/pkg/providers/internal/openstack/README.md
+++ b/pkg/providers/internal/openstack/README.md
@@ -63,10 +63,12 @@ for role in member load-balancer_member manager; do
 done
 ```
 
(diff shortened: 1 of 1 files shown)


squaremo (approved):
Yep, thank you.
Lines up with https://github.com/nscaledev/uni-python-unikorn-openstack-policy/pull/13 where the policy was changed. (**NB** the predicate `is_project_manager` there)

PR #493, opened 2026-04-28
Return MAC addresses from the servers V2 API

spjmurray:
## Summary
- add  to the servers V2 OpenAPI status schema and examples
- persist the primary server interface MAC address into server status from the provider layer
- return the stored MAC address from the V2 handler conversion and extend focused tests

## Validation
- make validate
- go test ./pkg/providers/internal/openstack -run 'TestReconcileServerPort|TestServerForCreate'
- go test ./pkg/handler/server


spjmurray:
Sadly we cannot test this in isolation, we need to do it via compute, so here's my local environment, end-to-end:

```
• Confirmed via the live compute API.

  Request used:

  - GET https://compute.spjmurray.co.uk/api/v2/instances

  Returned instance:

  - name: hilarious-dog
  - id: a200324d-375f-481d-b486-a6f685924106

  Returned status included:

  - macAddress: fa:16:3e:5f:91:4a

  Relevant response fragment:

  {
    "status": {
      "macAddress": "fa:16:3e:5f:91:4a",
      "networkId": "eaedfb5d-f626-42c6-a782-5ca484ad6818",
      "powerState": "Running",
      "privateIP": "192.168.0.166",
      "regionId": "f41d1700-a6d8-44e0-87eb-f88c9ee91002"
    }
  }

  So the MAC address is visible from the compute API for the existing instance.
```

mkjpryor (approved):

PR #496, opened 2026-04-30
Version v1.16.3

spjmurray:
The shared identity middleware now populates the propagated principal account type globally at the public API boundary from the authenticated request context, and the shared identity client injector forwards that principal automatically on downstream impersonated service-to-service calls.

`region` does not need to derive account type ad hoc in `InjectUserPrincipal()`. That helper only patches request-scoped tenancy fields such as organization and project IDs when those values are known from the current route or resource context but are not already present on the propagated principal.

This change updates `github.com/unikorn-cloud/identity` to the version that applies the new impersonation routing against the propagated principal type.

### DIFF
--- a/charts/region/Chart.yaml
+++ b/charts/region/Chart.yaml
@@ -4,8 +4,8 @@ description: A Helm chart for deploying Unikorn's Region Controller
 
 type: application
 
-version: 1.16.2
-appVersion: 1.16.2
+version: 1.16.3
+appVersion: 1.16.3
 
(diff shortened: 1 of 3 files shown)


spjmurray:
The review is assuming `region` must populate the propagated principal account type in `InjectUserPrincipal()`, but that is not where the field is derived.

`Principal.Type` is populated globally by the shared identity middleware at the public API boundary from the authenticated request context, and the shared identity client injector then propagates that principal automatically on downstream impersonated service-to-service requests.

`InjectUserPrincipal()` only patches request-scoped tenancy fields such as `organizationId` and `projectId` when those values are known from the current route or resource context but are not already present on the propagated principal.

PR #497, opened 2026-04-30
Remove restrictions around OS family and distribution in image metadata

yankcrime:
In the spirit of https://github.com/nscaledev/uni-specifications/pull/40, remove these restrictions from region machinery so that users can specify any string for the OS family and distribution in their image metadata.


squaremo (approved):
Seems OK to me. As a strict relaxation of constraints, I wouldn't expect this to cause backward-compability problems.

PR #505, opened 2026-05-07
Implement Octavia load balancer reconciliation

mkjpryor:
Adds OpenStack Octavia load balancer reconciliation, including client support, create/update, floating IP, and delete flows. Updates the load balancer provisioner, RBAC, CRD/OpenAPI generated artifacts, and API fixtures to expose lifecycle status consistently. Expands provider, provisioner, handler, and integration-style API coverage for load balancer lifecycle behavior. Validated with make touch, make license, make validate, make lint, make generate, a clean generated-code status check, and make test-unit.


pkg/handler/loadbalancer/client_v2.go
```diff
@@ -54,6 +54,8 @@ import (
 const (
 	endpoint = "region:loadbalancers:v2"
 
+	defaultIdleTimeoutSeconds = 60
```
spjmurray, line 57: You could just add a default to the CRD schema and the the K8S API fill it in for you... but then again there is already precedent will be your counter argument.  Meh.

spjmurray (commented):
```  
  Minor items worth considering
                                                                                                                                    
  1. effectiveMemberCount == 0 yield — add one sentence to the "Octavia Load Balancers" section of the OpenStack README noting that
  the provisioner yields until at least one member is registered, so an LB spec with empty pool members will never reach Available. 
  2. loadBalancerNetwork is resolved independently in both the provisioner and the provider — the provisioner needs it for readiness
   gating; the provider needs it to extract the subnet ID. The duplication is load-bearing, but a reader might wonder. The          
  provisioner README already implicitly covers this by listing "resolves the parent Network" as a separate step from provider
  reconciliation.                                                                                                                   
  3. //nolint:cyclop,gocognit on createLoadBalancer/deleteLoadBalancer — these are appropriate for orchestration code that
  necessarily has a non-trivial control flow. Not a concern.                                                                        
  
  ---                                                                                                                               
  Bottom line                                            
             
  The implementation is in keeping with the spirit and precedent of the design. It uses deterministic lookup over mirrored state,
  puts allocation release in the controller rather than the handler, uses sagas for compensatable writes, and populates the         
  documentation knowledge graph correctly. The branch is ready to merge from a design-consistency perspective.
```

I don't see anything visually, or with my buddy, that is too concerning.  You may want to consider point 1 above though, before I approve

pkg/handler/loadbalancer/client_v2.go
mkjpryor: I actually originally did do that, but timeout is not supported for UDP listeners. So you would have a UDP listener with 60s as the timeout that is silently ignored. I thought better to default to nil or not present and apply the default in code for the TCP protocol only.

spjmurray (approved):
Bless you my child

spjmurray (approved):

PR #509, opened 2026-05-08
Enforce region ACL on all user-supplied region IDs

spjmurray:
Previously, `FilterRegions` correctly gated which regions appeared in
list responses, but no equivalent check existed when a region ID was
supplied as input (URL path parameter, query parameter, or request body
field).  A caller who learned a restricted region's ID through any means
could pass it directly to create or read operations and have the service
act on their behalf against a region they were not authorized to use —
a classic confused-deputy privilege escalation.

This patch closes that gap by introducing two helpers in
`pkg/handler/region/region.go`:

- `checkAccess(ctx, *Region) error` — unexported, applies the same ACL
  rules as `FilterRegions` to an already-fetched region object.  Returns
  `HTTPNotFound` (not `HTTPForbidden`) so that unauthorized callers
  cannot distinguish between a region that does not exist and one they
  cannot see.

- `CheckAccess(ctx, regionID string) error` — exported, fetches the
  region by ID and delegates to `checkAccess`.  Used everywhere a
  user-supplied region ID first enters the handler or client layer.

`FilterRegions` is refactored to call `checkAccess` directly, keeping
the ACL logic in one place.

`GetDetail` calls `checkAccess` on the already-fetched object (no
second round-trip).  All other entry points call `CheckAccess`:

- `region.ListFlavors` and `region.ListExternalNetworks` (V1)
- `image.ListImages`, `image.CreateImage`, `image.DeleteImage` (V1)
- `image.QueryImages` (V2 — `GET /api/v2/regions/{regionID}/images`)
- `identity.generate` — covers both the user-facing `Create` path and
  the internal `CreateRaw` path used by the network saga; defense in
  depth since the saga's region ID is already validated at the network
  layer
- `network.CreateV2` — check added before the saga runs
- `storage.validateRegion` — upgraded from an existence-only check to a
  full ACL check; error message updated accordingly

Five new unit tests in `pkg/handler/region/region_test.go` cover the
cases: global (unrestricted) region, allowed org, denied org, global-
scope platform admin, and non-existent region ID.  The existing V2
image handler test is updated to seed the fake Kubernetes client with
the test region object so `CheckAccess` can resolve it.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>


squaremo (approved):
Approved on the basis of the explanation, having tests, and the automated review.

PR #515, opened 2026-05-12
feat(monitor): add scheduling duration metric and source provision duration from Nova timestamps

peterj-ns:
## Summary

- Replaces the annotation-based Pending-entry timestamp with \`OS-SRV-USG:launched_at\` from Nova, which is set at the exact moment the hypervisor boots the VM — independent of poll period
- Adds \`unikorn_region_server_scheduling_duration_seconds\` histogram measuring \`Uni CreationTimestamp → Nova created_at\` (time from CR creation until Nova accepted and processed the creation request, covering Uni reconcile overhead and Nova's initial scheduling work)
- Together the two histograms share the same origin (\`Uni CreationTimestamp\`); their difference (\`provision − scheduling\`) isolates Nova allocation + hypervisor boot time from Uni reconcile overhead
- Adds \`ScheduledAt\` and \`LaunchedAt\` to \`ServerStatus\`; both sourced from Nova on every provider observation
- Both metrics fire only once per server (nil→non-nil guard prevents duplicate observations on stop/restart cycles)
- Removes \`managePendingAnnotation\` and \`ServerPendingEntryTimeAnnotation\` entirely
- Updates \`unikorn_region_server_provision_duration_seconds\` description to accurately reflect what it measures

## Caveats

\`launched_at\` is when the hypervisor boots the instance, not when the guest OS finishes booting. For VMs this gap is negligible (<1 min); for baremetal it can be ~15 minutes. Closing it requires a guest-side signal and is out of scope here.

## Test plan

- [x] Unit tests updated and passing (\`make test-unit\`)
- [x] All lint/validate checks pass (\`make lint\`, \`make validate\`)
- [x] Generated deepcopy and CRD YAML regenerated and checked in
- [x] Verified Nova populates \`OS-SRV-USG:launched_at\` and \`created\` on real running servers in the dev cluster before deploying

Closes [INST-727](https://linear.app/nscale-workspace/issue/INST-727/observability-phase-4b-define-custom-otel-metrics-for-server-lifecycle).

### DIFF
--- a/pkg/apis/unikorn/v1alpha1/types.go
+++ b/pkg/apis/unikorn/v1alpha1/types.go
@@ -916,6 +916,9 @@ type ServerStatus struct {
 	PublicIP *string `json:"publicIP,omitempty"`
 	// MACAddress is the MAC address of the server's primary network interface.
 	MACAddress *string `json:"macAddress,omitempty"`
+	// LaunchedAt is the time Nova booted the VM, sourced from OS-SRV-USG:launched_at.
+	// Nil until the server has been observed Running at least once.
+	LaunchedAt *metav1.Time `json:"launchedAt,omitempty"`
 }
 
 // OpenstackServerList is a typed list of servers.
--- a/pkg/apis/unikorn/v1alpha1/types.go
+++ b/pkg/apis/unikorn/v1alpha1/types.go
@@ -916,9 +916,12 @@ type ServerStatus struct {
 	PublicIP *string `json:"publicIP,omitempty"`
 	// MACAddress is the MAC address of the server's primary network interface.
 	MACAddress *string `json:"macAddress,omitempty"`
-	// LaunchedAt is the time Nova booted the VM, sourced from OS-SRV-USG:launched_at.
-	// Nil until the server has been observed Running at least once.
+	// LaunchedAt is the time the provider booted the VM. Nil until the server has
+	// been observed Running at least once.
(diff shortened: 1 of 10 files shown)


pkg/apis/unikorn/v1alpha1/types.go
```diff
@@ -916,6 +916,13 @@ type ServerStatus struct {
 	PublicIP *string `json:"publicIP,omitempty"`
 	// MACAddress is the MAC address of the server's primary network interface.
 	MACAddress *string `json:"macAddress,omitempty"`
+	// LaunchedAt is the time Nova booted the VM, sourced from OS-SRV-USG:launched_at.
```
squaremo: Technically this could be another provider (i.e., not OpenStack), and we'd have to source it from that, and it may not be available from another provider's API. But I think we have just have a more neutral description, rather than not having this measurement.

squaremo (dismissed):
Let's run with this; thanks for your perseverance, Peter.

pkg/apis/unikorn/v1alpha1/types.go
peterj-ns: Yeah, I was thinking the same how much did we want to couple the observability to the underlying provider.

squaremo (approved):

pkg/apis/unikorn/v1alpha1/types.go
peterj-ns: Updated the description anyway to be more generic

PR #518, opened 2026-05-19
fix(deps): bump unikorn-cloud/core to v1.16.1 to unblock Server deletion

alex-emery:
## Summary

- Bumps `github.com/unikorn-cloud/core` from `v1.14.0-rc1.0.20260422145135-733e5169a07e` (2026-04-22) to `v1.16.1`.
- Picks up [unikorn-cloud/core #293](https://github.com/nscaledev/uni-core/pull/293) which makes `manager.RemoveResourceReference` (and its plural form) return `nil` on NotFound instead of propagating it.
- Fixes Server resources stuck in `Terminating` after their referenced SSH certificate authority has been deleted. Supersedes the heavier finalizer-hold approach in #516.

## Bug behaviour

A Server points at an SSH CA. While the Server is Running, the controller has released its finalizer on the CA (`removeConsumedResourceReferences` runs at the tail of `Provision`), so the CA is freely deletable. After the CA is deleted, the user deletes the Server. `Deprovision` calls `clearConsumedResourceReferences` → `removeSSHCertificateAuthorityReference` → `manager.RemoveResourceReference`. In the previously pinned uni-core version the keyed `Get` of the missing CA was propagated as a hard error; the manager reports `Reconciler error`, controller-runtime backs off, and the Server's finalizer is never removed. The user sees a Server stuck `Terminating` with status `Errored: ... not found: failed to clear SSH certificate authority reference`.

## Why the dep bump is the fix

The cleanup path in `pkg/provisioners/managers/server/provisioner.go` already does the right thing, but it relied on a tolerance in `manager.RemoveResourceReference` that only landed upstream in uni-core commit `50d0c2f` (PR #293, 2026-04-24) — two days *after* the previously-pinned commit (`733e5169`, 2026-04-22). No source changes are needed in this repo once the dependency catches up.

## Verification

- `make touch / license / validate / lint / generate / test-unit` all pass with only `go.mod` and `go.sum` modified.
- Live verified against a local Tilt cluster: a real Server that had been stuck `Terminating` for ~25 minutes under the old pin reconciled to `deletion complete` within ~1 s of the new pod taking the leader lease.

## Out of scope

`Provision`-side paths (`getSSHCertificateAuthority`, the singular `manager.AddResourceReference`) are still NotFound-intolerant. After a CA disappears the controller will still loop noisily until the Server is deleted — but deletion now works. A follow-up can decide whether to swallow NotFound there or treat the Server as Errored more gracefully. Not blocking on it here.

## Test plan

- [x] Unit tests pass (`make test-unit`).
- [x] Lint / generate / license / validate clean.
- [x] Live cluster: pre-bump, the stuck Server stays Terminating with `failed to clear SSH certificate authority reference` errors.
- [x] Live cluster: post-bump (Tilt rebuild), the same Server is garbage-collected within seconds.
- [x] (reviewer) Sanity-check uni-core changelog between `v1.14.0-rc1` and `v1.16.1` for anything else that affects region.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/spf13/pflag v1.0.10
 	github.com/spjmurray/go-util v0.1.3
 	github.com/stretchr/testify v1.11.1
-	github.com/unikorn-cloud/core v1.14.0-rc1.0.20260422145135-733e5169a07e
+	github.com/unikorn-cloud/core v1.16.1
 	github.com/unikorn-cloud/identity v1.14.0-rc1.0.20260430090955-4839bc57a519
 	go.opentelemetry.io/otel v1.43.0
 	go.opentelemetry.io/otel/sdk/metric v1.43.0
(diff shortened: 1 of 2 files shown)


spjmurray (approved):
Does what it says on the tin!

PR #521, opened 2026-05-19
test: fix skipped API coverage

sravanmedarapu-work:
## Summary
- Pass TEST_SECONDARY_PROJECT_ID through the API test workflow for Dev and UAT.
- Keep custom image IDs available for the already-deleted image assertion while making AfterAll tolerate not-found cleanup.
- Document the shared API test account mapping for Dev and UAT/QA token secrets.

## Why
Dev skipped coverage because TEST_SECONDARY_PROJECT_ID was not available, and the image delete flow cleared state before the follow-up not-found assertion. The README now records that the primary Dev and UAT/QA API test tokens are owned by qa-testaccount02@nscale.com so the GitHub variables and secrets are easier to map back to the test account.

## Validation
- go test -tags integration ./test/api/suites -run '^$'
- git diff --check


spjmurray (approved):
Let's make GREEN great again!

PR #523, opened 2026-05-20
Support instance rebuild on image change (INST-920)

alex-emery:
## Summary

Region-side support for INST-920 — an image change on an existing server now triggers an **in-place OpenStack Nova rebuild** instead of being silently dropped.

- Adds `RebuildServer` to the internal OpenStack `ServerInterface` (wraps gophercloud `servers.Rebuild`).
- `reconcileServer` compares the live Nova image against the desired image and rebuilds in place on a mismatch, yielding while the rebuild runs; a `REBUILD`-state server reports as provisioning rather than degraded.
- New shared image/flavor validator wired into the v2 create (POST) and update (PUT) handlers: image exists / visible / ready, flavor exists, disk large enough, virtualization compatible.

No API schema change. Pairs with the matching uni-compute PR — the two ship together.

## Test plan

- [x] `make test-unit` and the full pre-commit checklist (touch / license / validate / lint / generate / test-unit).
- [x] Live on a DevStack region: an image change drove an in-place Nova rebuild — the OpenStack server kept the same ID through ACTIVE -> REBUILD -> ACTIVE.
- [x] Live: an invalid image on an update was rejected by the API with 404 before reaching the controller.

## Known follow-up

A pre-existing delete/recreate name-collision race was observed during testing (a recreated server can adopt the previous, still-terminating Nova server by name). It affects name/flavor recreates and is not introduced or worsened by this change; tracked separately.


pkg/providers/internal/openstack/provider.go
```diff
@@ -2296,6 +2296,7 @@ func (p *Provider) reconcileLoadBalancerFloatingIP(ctx context.Context, client F
 // servers that already exist.
 func reconcileServerImage(ctx context.Context, client ServerInterface, server *unikornv1.Server, openstackServer *servers.Server) (*servers.Server, error) {
 	log := log.FromContext(ctx)
+	log.V(1).Info("server already exists")
```
squaremo: This is going to be pretty noisy. Do we need it, and if so, could it be at a higher verbosity level.

spjmurray (approved):
Approving — this has come a long way and it's a genuinely strong piece of work.

Both polish items from the last round are in, and the version-skew reachability catch on the pending tail (with `TestReconcileServerImageUnknownMarkerStateYields` to pin it) is a nicer fix than what I suggested. Everything from the original architectural pass is resolved: parallel retry engine gone, steady-state polling gone, monitor is the sole Nova poller feeding a single settlement wake, the copy-back/lost-health bug fixed, the transient-ACTIVE race closed via the ref∧task_state conjunction, and userdata-on-rebuild cleanly deferred with a documented contract note. CI is green across the board including the Integration suite and both contract directions.

Two things I'm signing off *with*, neither blocking: the Nova accept-atomicity behaviour rests on the live-region e2e probe rather than a gated CI check, and the rebuild state machine is dense enough that its load-bearing invariants (single-patch marker+health, the level predicate, task_state visibility) need respecting by whoever touches it next — now test-backed, so that's manageable.

The broader 'bounded-retry-before-terminal wants to be a core-owned disposition' discussion stays a separate upstream thread and isn't a condition of this merge.

Nice work @alex-emery.

PR #524, opened 2026-05-21
server uniqueness

spjmurray:
- **feat(server): replace hostname-uniqueness hack with deterministic UUIDs**
- **fix(server): harden deterministic ID generation and clarify uniqueness semantics**
- **feat(server): enforce name immutability on update with HTTP 422**


squaremo (commented):
Please remove the prefixes in all the commit messages.

squaremo (approved):
I agree with the machine ✅ 

You should give this a better PR title though, because that is what appears in the git history (and in any generated changelog).

PR #526, opened 2026-05-22
Fix network allocation leak

squaremo:
In end to end testing, we sometimes see left-over allocations. In at least one case it's caused by a test failing before the an implicitly-created OpenStack Identity is ready. The cleanup thunk deletes the Network, but this doesn't clean everything up.

This fix comes in three parts:
- refactor the provisioners, elaborating on the base.Base struct, so that the identity client (used for allocations) can be injected;
- add a test that the deletion does release the allocation;
- fix the underlying problem by only attempting to delete the OpenStack network if the identity is ready or it's already been created in OpenStack

Reversing the latter two steps shows that the test fails without the fix, and succeeds with it.

### DIFF
--- a/pkg/provisioners/managers/network/provisioner.go
+++ b/pkg/provisioners/managers/network/provisioner.go
@@ -88,6 +88,34 @@ func (p *Provisioner) Object() unikornv1core.ManagableResourceInterface {
 	return p.network
 }
 
+func identityReady(identity *unikornv1.Identity) bool {
+	condition, err := identity.StatusConditionRead(unikornv1core.ConditionAvailable)
+	if err != nil {
+		return false
+	}
+
+	return condition.Reason == unikornv1core.ConditionReasonProvisioned
+}
+
+func providerResourceIDsRecorded(network *unikornv1.Network) bool {
+	if network.Status.Openstack == nil {
+		return false
+	}
+
+	return network.Status.Openstack.NetworkID != nil ||
+		network.Status.Openstack.SubnetID != nil ||
+		network.Status.Openstack.VlanID != nil
+}
+
+func needsProviderDelete(identity *unikornv1.Identity, network *unikornv1.Network) bool {
+	// A v2 network can be deleted after the API has created its identity-side
+	// allocation but before provisioning has waited for Identity readiness and
+	// recorded provider resource IDs. In that window there is nothing for the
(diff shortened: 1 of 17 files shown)


pkg/provisioners/managers/network/provisioner.go
```diff
@@ -122,8 +143,10 @@ func (p *Provisioner) Deprovision(ctx context.Context) error {
 		return err
 	}
 
-	if err := provider.DeleteNetwork(ctx, identity, p.network); err != nil {
-		return err
+	if needsProviderDelete(identity, p.network) {
```
nsricardor, line 146: This still unconditionally calls ProviderAndIdentity before quota cleanup. Could it fail in the allocation-only cleanup path? If so, we may need to run the partial-state guard before that lookup.

squaremo: Hmm, good question. The create saga (in the handler) does this:
- validation
- create the identity ("service principal")
- create the allocation
- create the network Kubernetes object, for the provisioner to act on

That means it must have created the identity before it got to creating the allocation. There are other ways the `ProviderAndIdentity` call could fail, I suppose, but we need the identity to ... actually do we need the identity here? Or just the network, to see if it's got the fields that would indicate it's been created in OpenStack.

Have I understood your question at least :-)

nsricardor: Ok, I see the full picture now. In the normal create flow, the identity should always exist by the time the netwok is created. However, the delete path is less explicit as DeleteV2 in the handler deletes the identity and relies on cascading deletion to eventually delete the network, and it encodes an indirect lifecycle assumption identity existence is being used as a proxy for whether cleanup is needed.

That said, the fix looks like is doing what it's supposed to do. Tests are passing and the new test covers this scenario.

nsricardor (approved):

pkg/provisioners/managers/network/provisioner.go
squaremo: 👍 

Relying on cascading delete makes me nervous (and finalizers are a notorious pain point). The main weakness is that reasoning about deletion has to involve the whole, complicated model of what does what and when.

My imagination doesn't stretch to a simpler model (without replacing cascading delete entirely) here and now ... I anticipate that we'll have a few problems of this kind, so perhaps a pattern will emerge 🤞

PR #527, opened 2026-05-26
test(api): expand integration test coverage for LBs, networks, and auth

JakeNScale:
### Client fixes
- `GetNetwork`, `DeleteNetwork`, `GetLoadBalancer`, `DeleteLoadBalancer` now use `DoRequest(0)` + explicit
  status switch, matching the pattern used by the security group and SSH CA methods. 404 responses now
  return `ErrResourceNotFound` instead of `ErrUnexpectedStatusCode`.
- `WaitForLoadBalancerGone` updated to match the corrected error type.

### New tests
- **Auth**: unauthenticated requests to both the compute API and region service API return 401.
- **Load balancers**: not-found GET/DELETE, invalid body rejection, idempotent delete (repeated
  deletes return 202/404), UDP listener with no `idleTimeoutSeconds`, publicIP toggle (clears
  `status.publicIP`, preserves VIP address).
- **Networks**: not-found GET/DELETE, invalid body rejection, cross-org isolation (secondary org
  cannot access a network owned by the primary org).
- **Security groups**: invalid body rejection.

Closes INST-935

### DIFF
--- a/test/api/api_client.go
+++ b/test/api/api_client.go
@@ -494,21 +494,29 @@ func (c *APIClient) CreateNetwork(ctx context.Context, request regionopenapi.Net
 }
 
 // GetNetwork gets a specific network by ID.
+// Returns ErrResourceNotFound if the network does not exist.
 func (c *APIClient) GetNetwork(ctx context.Context, networkID string) (*regionopenapi.NetworkV2Read, error) {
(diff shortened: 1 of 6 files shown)


sravanmedarapu-work (approved):
LGTM

PR #534, opened 2026-05-27
Unblock deleting unprovisioned resources

squaremo:
Failed provisioning can leave Region resources with finalizers, before any recoverable OpenStack state exists. Delete paths should clear those control-plane objects instead of retrying client setup forever, while still propagating errors once provider state has been recorded.

### DIFF
--- a/pkg/providers/internal/openstack/provider.go
+++ b/pkg/providers/internal/openstack/provider.go
@@ -1331,6 +1331,18 @@ func (p *Provider) CreateIdentity(ctx context.Context, identity *unikornv1.Ident
 	return nil
 }
 
+func openstackIdentityHasComputeState(identity *unikornv1.OpenstackIdentity) bool {
+	return identity.Spec.SSHKeyName != nil || identity.Spec.ServerGroupID != nil
+}
+
+func canSkipOpenStackIdentityComputeDelete(err error) bool {
+	if gophercloud.ResponseCodeIs(err, http.StatusUnauthorized) {
+		return true
+	}
+
+	return kerrors.IsNotFound(err) || errors.Is(err, coreerrors.ErrConsistency)
+}
+
 // DeleteIdentity cleans up an identity for cloud infrastructure.
 //
 //nolint:cyclop,gocognit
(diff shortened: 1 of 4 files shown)


pkg/providers/internal/openstack/provider.go
```diff
@@ -1659,6 +1730,10 @@ func (p *Provider) CreateNetwork(ctx context.Context, identity *unikornv1.Identi
 	return nil
 }
 
+func networkHasOpenStackState(network *unikornv1.Network) bool {
+	return network.Status.Openstack != nil && network.Status.Openstack.NetworkID != nil
```
nsricardor, line 1734: All the other scenarios look solid. This one is the only case I’m not fully sure is reliable: reconcileLoadBalancer can create the Octavia LB, get back PENDING_CREATE with no VIP yet and yield (classifyOctaviaStatus fn)

If deletion happens in that window, we may skip cleanup even though the LB exists.

squaremo: Which one are you not sure about, network (the line commented upon), or load balancers (the func mentioned)? Or the interaction between the two...

squaremo: With respect to load balancers, we can't do any operations while it's in a PENDING_ state; it should yield and come back later to attempt the delete. This causes problems during test cleanup, which do things like:

 1. delete the load balancer
 2. delete other stuff

The tests bail on step 1. taking too long, and step 2 (some other cleanup) never happens. I'm fixing that in another commit.

nsricardor: Sorry, this comment was anchored in the wrong place. It was meant for loadBalancerHasOpenStackState.

The scenario is: if Octavia creates the LB, returns PENDING_CREATE with no VIP yet, and deletion then hits a skipped client setup error, we may remove the finalizer without ever calling GetLoadBalancer/DeleteLoadBalancer, leaving the LB behind.

The risk is “pending/no recorded status PLUS client setup failure.” That said, I wasn’t able to run this locally to test, but Claude confirmed the possibility.

PR #537, opened 2026-05-28
Cap file storage attachment parallelism to usable network range

nsricardor:
Updates file storage attachment handling so networks with smaller non-empty storage ranges are accepted and used in full, instead of failing when the range is smaller than the storage class parallelism target. Also updates API/CRD documentation and adds unit coverage for the capped range behavior.


pkg/handler/storage/client.go
```diff
@@ -372,10 +372,9 @@ func validateStorageRange(network *regionv1.Network, parallelism int) (*regionv1
 		big.NewInt(0).SetBytes(endIP),
 		big.NewInt(0).SetBytes(startIP),
 	)
-	required := big.NewInt(int64(parallelism - 1))
 
-	if available.Cmp(required) < 0 {
-		return nil, errors.HTTPUnprocessableContent("network storage range does not have enough available addresses for the requested parallelism")
+	if available.Sign() < 0 {
```
squaremo: This is correct but confusingly so. If I took at it face value, I would expect the comparison to be

    if available.Sign() <= 0 // if there's no IP addresses in the range, or the range is around the wrong way,  then there's a problem

because `0` available is also an error. It happens that the subtraction above will give one fewer than the count of the _inclusive range_ (which is what I'd call the available IPs), so this is effectively

    if (available - 1) < 0

I'd rather see the available count calculated correctly, then the comparison makes sense.

squaremo (dismissed):
Nice and tidy, thank you. If you think my comment on `available` is worth following up I'm happy to rubber stamp again :-)

pkg/handler/storage/client.go
nsricardor: Agreed, it was a bit confusing and I've updated the logic. Hopefully this is cleaner.

squaremo (approved):
Choice 🍎 Thanks Ricardo

PR #543, opened 2026-06-01
project observed FileStorage attachment status in API

nsricardor:
Updates FileStorage attachment status to project observed provisioning state and adds extensible mountOptions for NFS client hints. Currently emits remoteports, which lets clients mount across the observed storage IP range instead of a single endpoint for better NFS throughput/resilience.


pkg/handler/storage/client.go
```diff
@@ -128,16 +132,61 @@ func convertStatusAttachmentList(in *regionv1.FileStorage) *openapi.StorageAttac
...
+	case regionv1.AttachmentErrored:
+		return coreopenapi.ResourceProvisioningStatusError
+	case regionv1.AttachmentDeprovisioning:
+		return coreopenapi.ResourceProvisioningStatusDeprovisioning
+	default:
+		return coreopenapi.ResourceProvisioningStatusUnknown
+	}
+}
+
+func mountOptionsFromAttachmentStatus(in regionv1.FileStorageAttachmentStatus) *map[string]string {
```
squaremo, line 167: I guess this is a pointer to a map because of the wacky OpenAPI code generation, which doesn't understand maps are already nullable. OpenAPI code generation, sigh.

squaremo (approved):
Useful improvement and a good sized change, thanks Ricardo 🍏

PR #546, opened 2026-06-01
Add snapshot policies to file storage CRD

vinimdocarmo:
Adds the CRD-only storage model for inline FileStorage snapshot policies:
- Adds optional spec.snapshotPolicies with map-list semantics keyed by name.
- Adds schedule, retention, interval, weekday, and status Go types.
- Enforces CRD validation for max policy count, policy names, interval/weekday enums, UTC HH:MMZ times, monthly days 1..28, retention.keep >= 1, and interval-specific schedule rules.
- Adds status.snapshotPolicies entries keyed by name with optional generic conditions only.
- Regenerates deepcopy code and deployable CRD YAML.
- Adds storage/CRD behavior tests.
- Updates storage-model package documentation.


pkg/apis/unikorn/v1alpha1/types.go
```diff
@@ -989,6 +989,13 @@ type FileStorageSpec struct {
 	// Attachments are the network attachments for the storage.
 	Attachments []Attachment `json:"attachments,omitempty"`
 
+	// SnapshotPolicies are the named snapshot protection rules for the storage.
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=4
+	// +optional
+	SnapshotPolicies []FileStorageSnapshotPolicy `json:"snapshotPolicies,omitempty"`
```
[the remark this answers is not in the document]
vinimdocarmo: We won't be adding snapshot policies through the CRD/k8s route, so it should be fine. The API layer will be the one adding those.

pkg/apis/unikorn/v1alpha1/filestorage_snapshot_policy_test.go
nsricardor, line 1: I don’t think we currently have tests for the CRD types, and since this mostly verifies the type shape and generated stuff rather than any behavior, I’m not sure how much value it adds. @squaremo , do you have any guidance here?

vinimdocarmo: Happy to remove in favour of unit and integration tests that will come in the feature

squaremo: I don't see much value in tests that are verifying code generation (we have bigger problems if code generation doesn't work as expected -- we're very likely to get a non-compiling code or other breakages if that's the case).
The exception might be validation rules that are complicated enough that they are difficult to write and read in place. A golden test set can help make that sort of thing legible, and guard against mistakes when updating them.

vinimdocarmo: Agreed. I'll remove it
Thanks both!

vinimdocarmo: @nsricardor I removed the old test. I talked to @squaremo and we agreed on another approach which is to add some tests to guard the validation rules. That's now been added to the PR.

nsricardor (approved):
I validated the CRD changes against the required snapshot policy behavior, and everything looks good to me. The schedule and retention shape satisfies all requirements, so this gets my ✅.

That said, it may be worth getting a second approval from one of the senior UNI maintainers.

squaremo (approved):
Quick scan of the tests, looks good 👍

PR #550, opened 2026-06-03
Fix stale VLAN allocation cleanup by network ID

sravanmedarapu-work:
## Context

Stage UAT found stale VLAN allocation rows where the Region `Network` CR had already been deleted, but `VLANAllocation` still held the VLAN entry. Confirmed examples from the investigation:

- `607 / def703a8-e37c-4264-a503-69532663149b`
- `608 / cc69f439-1afc-43a3-80f7-301ee64ef0de`
- `605 / c3a541a0-6938-4c6f-b4b8-00b3afd73344` remains unmapped but has the same stale allocator shape

Investigation doc: https://www.notion.so/nscalecloud/VLAN-605-607-608-Stale-Allocation-Investigation-uni-region-LoadBalancer-UAT-2026-06-03-374caf6bfadc8174b03afda2de7e0ae3

## Problem

The allocator records ownership as `VLANAllocationEntry.NetworkID`, but network delete previously freed VLANs by VLAN ID derived from the live OpenStack provider network or observed Network status. That can miss cleanup when:

- the OpenStack provider network is already gone, and
- Network status did not record the VLAN ID or is otherwise stale.

That leaves allocator state stale and can later surface as:

```text
allocation failure: vlan ids exhausted
```

## Changes

- Make VLAN allocation idempotent by `networkID`; repeated allocation for the same network returns the existing VLAN.
- Guard against duplicate existing allocations for the same `networkID`.
- Add `FreeByNetworkID`, which frees allocator entries by `NetworkID` and absorbs missing `VLANAllocation` objects/missing entries.
- Use `FreeByNetworkID(ctx, network.Name)` during OpenStack network delete, so cleanup does not depend on Network status or a live OpenStack network.
- Stop rolling back VLAN allocation after OpenStack network creation failure.
- Keep searchable logs for VLAN allocation failure and OpenStack network creation failure.
- Keep the testable delete helper and initialize the test provider's VLAN allocator consistently with the real provider.

## Verification

```text
go test ./pkg/providers/allocation/vlan
go test ./pkg/providers/internal/openstack
make lint
```

Passed locally.

### DIFF
--- a/pkg/providers/allocation/vlan/README.md
+++ b/pkg/providers/allocation/vlan/README.md
@@ -38,8 +38,10 @@ and the `Region` VLAN/provider-network configuration that this allocator uses.
 - `Allocate()` is idempotent by `networkID`; repeated allocation for the same
   network returns the existing VLAN ID.
 - At most one VLAN ID may be allocated to a given `networkID`. `Allocate()`
-  returns an error if duplicate allocations are detected for the same network,
-  while `FreeByNetworkID()` removes all matching entries to heal corruption.
+  prevents new duplicate ownership by returning an error if duplicates are
+  detected for the same network. `FreeByNetworkID()` deliberately removes all
+  matching entries when cleaning up a network so existing duplicate ownership is
+  healed during deletion.
 - `Free()` and `FreeByNetworkID()` are idempotent.
 - If the allocation table contains the same VLAN ID more than once, the package
   treats that as corruption and returns an allocation error rather than trying
--- a/pkg/providers/allocation/vlan/README.md
+++ b/pkg/providers/allocation/vlan/README.md
@@ -35,7 +35,12 @@ and the `Region` VLAN/provider-network configuration that this allocator uses.
   topology-aware.
 - Persistence and concurrency coordination rely on the backing Kubernetes
   `VLANAllocation` object rather than in-memory state.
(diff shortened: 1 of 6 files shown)


JakeNScale (dismissed):

spjmurray (changes requested):
Good crack, it does make it less likely to happen, but not impossible.  Here's my reasoning...

At present the VLAN in the network resource is observational, not operational, it's best effort, it's entirely possible to provision a network but fail to update the resource.  If you deleted it at that point you'd correctly delete the OS network by name, but the VLAN wouldn't be set and we'd not free it.  This is the inherent risk of drift/caching and why we avoid it e.g. why I was using the OS segmentation ID as the source of truth.

You are right in one respect though, just make the deallocation
* idempotent if it is not
* unconditional

Then the correct (imo) and infallible thing to do is
* change the VLAN allocation to allow only one per-network (add a guard in), that prevents accidentially allocating more than one
* don't attempt to deallocate it when provisioning fails, I've seen a number of occasions when it's due to someone has manaually allocated the VLAN for themselves via a back channel and that causes provisioning failure, putting it back in the pool just causes the problem to pass from one network to another
* do the free based on network ID, that's always present

spjmurray (approved):

PR #551, opened 2026-06-04
test(provisioners): unit test uplift for manager provisioners

JakeNScale:
Adds unit coverage for the three under-tested provisioner manager packages, following the `network` provisioner test pattern (`NewForTest` factory, gomock `Provider`/`Providers`, `coreclient.NewContext`, `fake.NewClientBuilder`).

## Coverage
- `security-group`: 0% → 68.8%
- `identity`: 0% → 63.6%
- `load-balancer`: 19% → 70.0%

## What's covered
- **security-group**: provision/deprovision lifecycle, unready identity inhibits provisioning, provider error propagation.
- **identity**: provision/deprovision lifecycle, reference and provider error propagation, reference is not removed when the provider delete fails.
- **load-balancer**: provider create error propagation, allocation cleanup on deprovision, idempotent cleanup when the allocation is already gone (404).

## Production change
The identity provisioner constructed `identityclient.NewReferences(...)` inline inside `Provision`/`Deprovision`, leaving no seam to exercise the lifecycle without real HTTP. Extracted a single `ReferenceManager` interface, default-constructed in `New` exactly as before (`*identityclient.References`). Production behaviour is unchanged — construction simply moves from per-call to `New`, matching how `network`/`load-balancer` build their identity client. Removed the now-unused `options` field to satisfy `staticcheck`.

## Notes
- `Provision … allocation rollback` (from the issue) does not exist in the LB code path — `Provision` propagates the provider error and the controller retries; allocation cleanup lives in `Deprovision`. Tests cover the real behaviour.
- The identity reference seam has no generated mock, so it uses a small hand-rolled fake that records call counts (enables the "reference not removed on provider-delete failure" ordering assertion); the provider boundary uses gomock.

Closes INST-950


pkg/provisioners/managers/load-balancer/provisioner_test.go
```diff
@@ -221,7 +232,94 @@ func TestDeprovision_ProviderDeleteFirst(t *testing.T) {
...
+
+	provider := mocktypes.NewMockProvider(ctrl)
+	provider.EXPECT().DeleteLoadBalancer(gomock.Any(), gomock.Any(), lb).Return(nil)
+
+	providers := mockproviders.NewMockProviders(ctrl)
+	providers.EXPECT().LookupCloud(testRegionID).Return(provider, nil)
+
+	mockIdentity := identitymock.NewMockClientWithResponsesInterface(ctrl)
+	mockIdentity.EXPECT().
+		DeleteApiV1OrganizationsOrganizationIDProjectsProjectIDAllocationsAllocationIDWithResponse(gomock.Any(), testOrganization, testProject, testAllocationID).
```
sravanmedarapu-work, line 316: @claude suggest short names for `DeleteApiV1OrganizationsOrganizationIDProjectsProjectIDAllocationsAllocationIDWithResponse`

sravanmedarapu-work (approved):
LGTM

PR #555, opened 2026-06-07
Deflake securitygroup PUT/GET test

squaremo:
The API handler reads through the controller-runtime cache, so a GET immediately after a successful PUT can still observe the previous security group spec.

Poll the integration assertion until the cache has seen the patched object while keeping the update response check synchronous.


spjmurray (approved):
Eventual consistency... shame it's not a write through cache

PR #559, opened 2026-06-09
fix: prevent VLAN leak on network create/delete failure

sravanmedarapu-work:
## Summary

Fixes allocation cleanup around network create/delete failure paths so allocation state stays consistent with network state.

The key behavior change is conservative ownership handling: an allocation is only released after the network is known to be absent or successfully deleted. If the operation outcome is ambiguous, the allocation is preserved to avoid handing the same identifier to another network.

## Changes

- Delete path: delete the network before releasing the corresponding allocation.
- Create path: when network creation fails after an allocation was reserved, verify the network is absent before releasing that allocation.
- Tests: add focused unit coverage for failed delete, successful delete, failed create with absent network, failed create with committed network, ambiguous confirmation, and paths where no allocation is reserved.

## Verification

- `go test ./pkg/providers/...`
- `golangci-lint run ./...`

## Notes

Incident-specific evidence, environment cleanup details, and resource identifiers are tracked internally rather than in this public PR.


pkg/providers/internal/openstack/provider.go
```diff
@@ -1484,6 +1484,25 @@ func storageRange(prefix net.IPNet, reservations *unikornv1.NetworkReservations)
 	}
 }
 
+// tryFreeVLANOnCreateFailure frees the VLAN allocation only when it can confirm
+// that Neutron did not commit the provider network. A transport or context error
+// may fire after Neutron has already created the network; freeing in that case
+// would allow the same VLAN ID to be handed to a different network and produce
+// VlanIdInUse on the next provision.
+func (p *Provider) tryFreeVLANOnCreateFailure(ctx context.Context, client NetworkInterface, network *unikornv1.Network, vlanID *int) {
```
spjmurray: I believe we decided to keep VLANs allocated for the lifetime of the network, freeing on delete.  This was a pragmatic decision in case someone had manually allocated it (returning to the pool would just cause the next network that happened to allocate it to fail) or some other problem that would be best left until it could be debugged.  I bring this up because it's a recurring problem and not freeing is the better UX decision.

spjmurray (commented):
I'll accept the reordering, the free on create fail I have some insight...

pkg/providers/internal/openstack/provider.go
sravanmedarapu-work: Got you, reverted.

spjmurray (approved):
This is fine!

PR #561, opened 2026-06-12
feat: expose provider provisioning status on servers

alex-emery:
## Summary

Surfaces baremetal lifecycle distinctions (queued / building / running) to
the user via the live `Phase` field, fed from Nova + Ironic state by the
existing monitor.

## Design

Per spjmurray's follow-up review on this PR, the original
`ProviderProvisioningStatus` override approach was solving the right
problem in the wrong field. `metadata.provisioningStatus` is
provisioner-owned ("did the CR provision") and a one-shot signal. Making
the monitor a second writer caused a `provisioned -> queued` regression
flip during normal baremetal create:

1. `CreateServer` returns nil as soon as Nova accepts → core reconciler
   sets `ConditionAvailable = Provisioned` → API reports `provisioned`
2. First monitor poll sees BUILD + pre-deploy Ironic state → override
   flips to `queued` → API regresses

Live state belongs on `Phase` (already monitor-owned, already on the API
surface as `Status.PowerState`). This PR extends the existing
`InstanceLifecyclePhase` enum:

- `Pending` (unchanged) — initial, not yet observed
- `Queued` (new) — provider accepted, hardware not yet engaged
  (baremetal pre-deploy Ironic states)
- `Building` (new) — provider actively working (VM `BUILD`, baremetal in
  deploy states, baremetal Ironic failure states)
- `Running` / `Stopping` / `Stopped` (unchanged)

The Ironic lookup machinery previously feeding the override now feeds
`setServerPhase`. Failure signals stay on `ConditionHealthy` (existing
behaviour); no `Errored` phase is introduced.

## UX contract change

**API consumers should treat `Phase=Running` as the readiness signal**,
not `provisioning_status=provisioned`. Once a server reaches
`provisioned`, Phase is the live indicator (`Queued` -> `Building` ->
`Running`). uni-ui and nscale-cli will need follow-up changes — flagged
but not blocking on this PR.

## Dependencies

- Closes the dependency on nscaledev/uni-core#305 (now closed); bumps
  `unikorn-cloud/core` to released `v1.17.1`.
- Consumer PR: nscaledev/uni-compute#299 drops its mirror of the override
  field and surfaces the new Phase values through the existing
  `PowerState` projection.


spjmurray (approved):
AI says merge

PR #562, opened 2026-06-15
Pin core schema reference to v1.17.0

spjmurray:
Previously the $ref URLs in server.spec.yaml and the import mapping in
config.yaml pointed at the main branch of unikorn-cloud/core. This caused
schema.go (which embeds the resolved spec) to change whenever core's main
branch moved, producing spurious diffs in unrelated pull requests. Pinning
to v1.17.0 keeps generated output stable until we deliberately bump the
version.


alex-emery (dismissed):
instructions to update the spec feel a bit hidden but otherwise LGTM

PR #564, opened 2026-06-16
Make provider deletes idempotent and identity-tolerant

spjmurray:
Commit a618471 ("Fix network allocation leak") fixed a real symptom the
wrong way. A v2 network can be deleted while its hidden service-principal
identity is still provisioning (OpenstackIdentity has no ProjectID yet).
DeleteNetwork then failed building its project-scoped client, the error
propagated out of Deprovision, and the downstream identity-side allocation
cleanup never ran, leaking quota.

That commit gated the provider delete behind

    needsProviderDelete = identityReady(identity) ||
                          providerResourceIDsRecorded(network)

where providerResourceIDsRecorded reads Status.Openstack.*. This went wrong
on two counts:

- It made a delete decision depend on best-effort, non-authoritative status
  (CreateNetwork even resets Status.Openstack to empty on every reconcile).
  A lost status write skips a delete that is genuinely required and re-leaks
  the resource. This contradicts the provider contract in
  pkg/providers/README.md: re-find backing resources rather than trusting
  mirrored provider-state records.
- It gated the delete on parent identity readiness, the exact "broad
  prerequisite" the commit's own new Deprovisioning Partial State note in
  pkg/provisioners/README.md said not to gate on.

The delete path was already designed to not trust status: it rediscovers
Neutron resources by name and frees the VLAN by network name, and is fully
idempotent. Gating it added risk and no benefit. The only real precondition
is being able to build a project-scoped client, which needs the identity to
be realized.

Fix it at the layer that owns the truth, and apply it everywhere, since the
same fragility existed unguarded in the other identity-consuming deletes
(DeleteSecurityGroup even dereferenced a nil ProjectID and panicked):

- Add openstackIdentityProvisioned and have DeleteNetwork, DeleteSecurityGroup,
  DeleteServer and DeleteLoadBalancer no-op when the identity was never
  realized (OpenstackIdentity absent or no ProjectID). Finalizer ordering
  guarantees the identity outlives its consumers, so at delete time it is
  either realized-and-complete or never realized; in both cases an
  unconditional idempotent delete is correct.
- Remove needsProviderDelete from the network provisioner; Deprovision now
  delegates unconditionally. The create-side ResourceReady requeue gate stays.
- Correct the docs that encoded the flawed assumption (provisioners and
  network READMEs) and add the idempotent/identity-tolerant delete invariant
  to the providers README.
- Cover the no-op paths for all four deletes, including the security group
  panic regression, and update the network provisioner test to expect the
  unconditional delete.

### DIFF
--- a/pkg/providers/README.md
+++ b/pkg/providers/README.md
@@ -58,6 +58,19 @@ packages are the concrete provider implementations.
     the real backing resources
   - mirrored provider-state records are not the preferred answer unless the
     state cannot be reconstructed safely enough by other means
+- Provider `Delete*` methods must be idempotent and must tolerate an unrealized
+  identity as a no-op. Callers delegate unconditionally; the provider
+  self-gates on realized identity rather than the caller gating on readiness or
+  status:
+  - rediscover backing resources by name rather than trusting recorded status,
+    and treat already-absent resources as success
+  - when the backing service-principal identity has not been realized (the
+    `OpenstackIdentity` is absent, or has no project allocated yet) nothing
+    could have been created, so the delete returns cleanly instead of erroring
+  - this is safe because finalizer ordering keeps the identity alive until its
+    consumers are gone: at delete time it is either realized-and-complete or
+    never realized. Callers must therefore never gate a delete on identity
+    readiness or recorded status — that belongs to this layer.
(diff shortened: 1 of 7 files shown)


sravanmedarapu-work (approved):
LGTM

PR #565, opened 2026-06-17
Use typed resource IDs across the region API

spjmurray:
Region's user-facing resource identifiers (region, identity, network, security
group, load balancer, server, SSH certificate authority, file storage, image,
flavor) are UUIDs, but they were passed around as bare strings: a malformed ID
was only caught — if at all — deep in a handler or at the Kubernetes client, and
the type system did nothing to stop an organization ID being used where a
network ID was meant.

This adopts identity's scope-reader surface and threads distinct, UUID-backed
typed IDs through the stack, applying one rule everywhere rather than rederiving
it case by case: preserve type safety until you can't — validate an ID at the
trust boundary, carry it typed, and convert to a string only at a genuine sink
(a Kubernetes object name or label, a CRD spec field, or a provider/region API).

End to end:

- Define UUID-backed typed IDs in pkg/ids and apply them in the OpenAPI spec via
  x-go-type, so non-UUID path parameters are rejected at the router (400) before
  any handler runs.
- Thread the typed IDs through every v1 and v2 handler client method (get,
  create, update, delete, the server lifecycle actions, the region client and
  the identity/console-session helpers). The by-name lookup primitives
  (GetRaw/GetV2Raw) stay string-keyed and convert at the ObjectKey, since they
  are also reached with stored label and CRD-spec strings.
- Consume identity as a black box for authz, tenancy and principal: RBAC via the
  scope-reader variants, tenancy via ids.OwnedByProject (replacing the stringly
  AssertProjectOwnership), principal attribution via the enrichment helpers
  (EnrichUserPrincipalProjectScope/...ID), and placement via
  SetIdentityMetadataProjectScope — removing region's bespoke principal
  injection and the last WithOrganization placement sites.
- Validate the IDs that feed security-relevant lookups: the create-body regionId
  (which gates private-region access through CheckAccess) and networkId (which
  selects the parent network) are typed, so they are UUID-validated at unmarshal
  instead of arriving as unchecked strings.

This settles two deliberate seams. Identity owns authorization, tenancy and
principal; region consumes that surface as a black box rather than re-deriving
it. And the typed IDs run from the router to the edge of the handler layer,
converting to strings only at the sinks (Kubernetes object names and labels, CRD
spec fields, provider/region calls); those sinks are a boundary, not a dead end
— the CRD and provider layers behind them can be given typed IDs in a later pass
without disturbing the handler surface settled here.

Deliberately left as strings: read-model and status IDs, and the storage
attachment NetworkIDList. These are populated from stored labels in the convert
functions, so typing them would force fail-closed parsing across the read path
for no input-validation gain (storage attachments are validated up front in
validateAttachments instead).

Integration fixtures and unit tests use UUID identifiers and assert the
security-correct status per condition (malformed -> 400, unauthorized
organization -> 403, private region accessed by a non-owner -> 404).

Built against the identity v1.17.7 release, which ships the scope-reader surface
this depends on.


pkg/client/references.go
```diff
@@ -51,6 +52,11 @@ func (r *References) httpClient(ctx context.Context, client client.Client, resou
 }
 
 func (r *References) AddReferenceToNetwork(ctx context.Context, networkID string, resource client.Object) error {
```
spjmurray, line 57: In the identity change we (should have) defined new type safe, backwards compatible variants for the analogous package. Do the same here

pkg/handler/identity/client.go
```diff
@@ -128,7 +129,7 @@ func (c *Client) generate(ctx context.Context, organizationID, projectID string,
 	}
 
 	out := &unikornv1.Identity{
-		ObjectMeta: conversion.NewObjectMetadata(&request.Metadata, c.Namespace).WithOrganization(organizationID).WithProject(projectID).WithLabel(constants.RegionLabel, request.Spec.RegionId).Get(),
+		ObjectMeta: conversion.NewObjectMetadata(&request.Metadata, c.Namespace).WithOrganization(organizationID.String()).WithProject(projectID.String()).WithLabel(constants.RegionLabel, request.Spec.RegionId).Get(),
```
spjmurray: Still uses WithOrganization, not the new identity specific constructors

pkg/handler/identity/client.go
```diff
@@ -154,20 +155,20 @@ func (c *Client) GetRaw(ctx context.Context, organizationID, projectID, identity
 		return nil, fmt.Errorf("%w: unable to lookup identity", err)
 	}
 
-	if err := coreutil.AssertProjectOwnership(resource, organizationID, projectID); err != nil {
+	if err := coreutil.AssertProjectOwnership(resource, organizationID.String(), projectID.String()); err != nil {
```
spjmurray: This is a seam that needs killing, discuss it with me

pkg/handler/identity/client.go
```diff
@@ -154,20 +155,20 @@ func (c *Client) GetRaw(ctx context.Context, organizationID, projectID, identity
...
 
 // List returns an ordered list of all resources in scope.
-func (c *Client) List(ctx context.Context, organizationID string) (openapi.IdentitiesRead, error) {
+func (c *Client) List(ctx context.Context, organizationID identityids.OrganizationID) (openapi.IdentitiesRead, error) {
 	var result unikornv1.IdentityList
 
 	options := &client.ListOptions{
 		LabelSelector: labels.SelectorFromSet(map[string]string{
-			coreconstants.OrganizationLabel: organizationID,
+			coreconstants.OrganizationLabel: organizationID.String(),
```
spjmurray: Likewise this appears to be an identity specific seam we can fold into a library call

pkg/handler/loadbalancer/client_v2.go
```diff
@@ -596,16 +601,16 @@ func generateListenerListV2(in []openapi.LoadBalancerListenerV2) ([]regionv1.Loa
...
 	listeners, err := generateListenerListV2(request.Spec.Listeners)
 	if err != nil {
 		return nil, err
 	}
 
 	out := &regionv1.LoadBalancer{
 		ObjectMeta: conversion.NewObjectMetadata(&request.Metadata, c.Namespace).
-			WithOrganization(organizationID).
-			WithProject(projectID).
+			WithOrganization(organizationID.String()).
```
spjmurray: Still calls WithOrganization etc

pkg/handler/loadbalancer/client_v2.go
```diff
@@ -755,10 +765,12 @@ func (c *Client) CreateV2(ctx context.Context, request *openapi.LoadBalancerV2Cr
...
 
-	organizationID := networkResource.Labels[coreconstants.OrganizationLabel]
-	projectID := networkResource.Labels[coreconstants.ProjectLabel]
+	organizationID, projectID, err := networkResource.OrganizationAndProjectID()
+	if err != nil {
+		return nil, err
+	}
 
-	if err := rbac.AllowProjectScopeCreate(ctx, c.Identity, endpoint, identityapi.Create, organizationID, projectID); err != nil {
+	if err := rbac.AllowProjectScopeCreateID(ctx, c.Identity, endpoint, identityapi.Create, organizationID, projectID); err != nil {
```
spjmurray: Can this not use a Reader variant?

pkg/handler/network/client.go
```diff
@@ -170,7 +171,7 @@ func (c *Client) generate(ctx context.Context, organizationID, projectID, identi
 	}
 
 	out := &unikornv1.Network{
-		ObjectMeta: conversion.NewObjectMetadata(&request.Metadata, c.Namespace).WithOrganization(organizationID).WithProject(projectID).WithLabel(constants.RegionLabel, identity.Labels[constants.RegionLabel]).WithLabel(constants.IdentityLabel, identityID).Get(),
+		ObjectMeta: conversion.NewObjectMetadata(&request.Metadata, c.Namespace).WithOrganization(organizationID.String()).WithProject(projectID.String()).WithLabel(constants.RegionLabel, identity.Labels[constants.RegionLabel]).WithLabel(constants.IdentityLabel, identityID).Get(),
```
spjmurray: This again needs to use the identity versions...

alex-emery (approved):

PR #566, opened 2026-06-19
Introduce Server create gates

squaremo:
For integration with external systems, it's useful to gate server creation on third-party status (like availability or health metrics), or just-in-time provisioning (like programming a network fabric). Following the design given in https://github.com/nscaledev/uni-specifications/pull/43, this change adds a creation gate mechanism to servers.


pkg/apis/unikorn/v1alpha1/helpers.go
```diff
@@ -253,6 +253,74 @@ func (r ActiveConditionReason) Message() string {
...
+	}
+
+	return out
+}
+
+func (c *Server) ProviderCreateGatesReady() bool {
+	return len(c.RemainingProviderCreateGates()) == 0
+}
+
+func (c *Server) ProviderCreateGatesReset(actor, reason, message string) {
```
alex-emery, line 318: `ProviderCreateGatesReset` has no production callers — it is defined and unit-tested, but nothing invokes it. Per the spec ("Region owns reset semantics"), its home should be the from-scratch create retry path: when `deleteFailedProviderServer` clears runtime status ahead of a new provider create attempt, the gates should be reset to `Unknown` so external services re-run their pre-create work. The unit test even uses reason `"ProviderCreateRetry"`, which suggests that wiring was intended. As it stands, a retried create sails straight through the gate check with the stale `True` statuses. Suggest either calling this from the retry path in this PR, or dropping the helper until the follow-up that wires it in.

alex-emery (approved):

PR #568, opened 2026-06-22
Openstack placement preflight for pinned servers

squaremo:
When booting a server that's pinned to a specific bare-metal machine, a fair proportion of the time the server you're targeting is in some unavailable state, like cleaning. (When not pinned, Nova is free to choose any available machine, so this is much less likely to happen). At present this means it fails to boot, and the server goes into an error state, and needs human intervention to recover.

This change introduces a "pre-flight check" when the server is pinned, run just before the server is created. The check asks the Nova placement service whether the specific machine is available, and if not, rather than trying the create request, it yields (so as to be retried later). There is a small time-of-use, time-of-check window here of course; but, this should nonetheless cut down on the number of servers that end up in an error state.


spjmurray (approved):
AI says this does what it should and also doesn't get in the way of the UX goals, so off we go!

PR #569, opened 2026-06-23
Retry server creation on error

squaremo:
Related to #568, another mitigation for transient conditions. In this case, experience shows that when a (Nova) server is in an error state, deleting it and creating again will often get past problems. Unlike #568, this applies to VMs, baremetal, and pinned baremetal -- though the latter two categories are more likely to end up in error states.

Also unlike #568, which does a pre-flight check specific to OpenStack, this can be implemented generically in the controller, rather than behind the provider abstraction. "Error" is kept as a terminal state: the controller will only mark a Server as being in error state once it's used up the retries. And, since there's no external signal otherwise, an event is recorded (and a log line logged) on each retry.

### DIFF
--- a/charts/region/crds/region.unikorn-cloud.org_servers.yaml
+++ b/charts/region/crds/region.unikorn-cloud.org_servers.yaml
@@ -241,6 +241,17 @@ spec:
               privateIP:
                 description: PrivateIP is the private IP address of the server.
                 type: string
+              providerCreateFailures:
+                description: |-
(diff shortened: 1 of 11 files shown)


spjmurray (approved):
Does what I as going to do.  Perhaps we may have to make a few tweaks to get the full end-to-end UX, and maybe clean up the semantics of who owns write permissions on what, but a tomorrow problem!

PR #570, opened 2026-06-24
Push typed resource IDs into CRD specs and the provider layer

spjmurray:
The previous pass threaded typed, UUID-backed IDs from the router to the edge of
the handler layer, converting to strings at the sinks: Kubernetes object names
and labels, CRD spec fields, and provider/region calls. That commit called those
sinks a boundary, not a dead end. This is the later pass it anticipated: it takes
the Server vertical past the CRD-spec and provider sinks so the same one rule —
preserve type safety until you can't — holds all the way down to the external
cloud SDK.

CRD spec fields:

- Type the region-owned UUID fields on the Server CRD: ServerSpec.FlavorID
  (FlavorID), ServerImage.ID (ImageID), ServerSecurityGroupSpec.ID
  (SecurityGroupID), and ServerNetworkSpec.ID (NetworkID). The last is set from
  the parent network's name and looked up as a region network, so the old
  "physical network ID" comment was wrong and is corrected to "region network ID".
- pkg/ids types are uuid.UUID, i.e. [16]byte. controller-gen builds the CRD
  schema statically from the Go type and never runs the marshaller, so a naive
  swap would emit an integer array. Add +kubebuilder:validation:Type=string (and
  Format=uuid) to each type declaration; the regenerated CRD is type: string,
  format: uuid with no array leak. Runtime round-trip already rides the existing
  TextMarshaler via the apimachinery codec, so no MarshalJSON is needed.
- Server.FlavorID() is dropped (callers read the typed spec field directly);
  Server.ImageID() keeps only its nil-image guard now that the ID is typed.

Provider layer:

- Type the Image surface, the only provider interface that passed bare ID
  strings: GetImage/DeleteImage take regionids.ImageID, and GetImage plus the
  AvailableToOrganization/OwnedByOrganization query predicates take
  identityids.OrganizationID. The object-passing interfaces (network, security
  group, load balancer, server, identity) need no signature change — they
  already carry their IDs inside the CRD object, which is now typed.
- The openstack and simulated providers convert to strings only where they hand
  an ID to an external system (gophercloud image cache/service, organization
  tags on provider images) or to a by-name Kubernetes lookup (lookupNetwork,
  lookupSecurityGroup, GetV2Raw). isBaremetalFlavor and the health monitor
  likewise convert at their string-config comparison.

The conversion functions split cleanly along the existing read-model boundary:
typed FlavorId/ImageId pass straight through, while the still-stringly read-model
fields (ServerNetwork.Id, ServerSecurityGroup.Id, the v2 status NetworkId) take
.String(). The v2 create path gains fail-closed UUID validation where it
previously accepted unchecked strings (the security-group list and the network
reference), matching the create-body validation philosophy of the prior pass.

Deliberately left as strings: the GetLabels() foreign-key reads that resolve a
resource's parent identity/region (base.go, the provisioners, the read-model
status). Those labels were designed for selector/ArgoCD discrimination, not data
carriage; promoting the parent reference to a typed spec field or owner reference
is a separate data-model migration with a backfill, not part of this typing pass.
Provider-internal and external identifiers (physical network, OpenStack flavor
metadata, node-class, vlan/subnet/router IDs) stay strings: they live in another
system's namespace, not region's.

Fixtures and unit tests use UUID identifiers throughout, replacing non-UUID
shorthands (metal, vm, org-1, image1). The reflect-mode ImageQuery mock under
pkg/handler/image/mock is regenerated for the typed predicate signature.


spjmurray (commented):
In general the smells I am seeing are related to parsing something that should have already been parsed/propagated.

alex-emery (approved):

PR #572, opened 2026-06-25
Poll read-after-write assertions against the informer cache

spjmurray:
The region API server serves single-resource GETs and Lists from the
controller-runtime cache, which is repopulated asynchronously by a watch.
Writes are synchronous to the apiserver, so an immediate read-back after a
create or update races the cache and can observe the pre-write object (a
stale name/spec, or a resource not yet present in a list).

This caused the networks "should reflect updated name and description"
spec to flake: the PUT response correctly echoed the updated name, but the
immediate round-trip GET returned the pre-update name.

Wrap every immediate read-back of a just-written resource in a bounded
Eventually (5s timeout, 250ms polling), matching the existing convention in
the load balancer and security group suites. The PUT/POST response
assertions remain immediate and authoritative; only the cache-backed
re-reads tolerate convergence.

### DIFF
--- a/test/api/suites/filestorage_test.go
+++ b/test/api/suites/filestorage_test.go
@@ -88,34 +88,38 @@ var _ = Describe("File Storage Management", func() {
 
 				GinkgoWriter.Printf("Created test storage for list: %s (%s)\n", testStorageName, testStorageID)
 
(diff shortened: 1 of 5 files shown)


sravanmedarapu-work (approved):
LGTM

PR #573, opened 2026-06-25
Checkout deployed refs for dev and UAT API tests

sravanmedarapu-work:
## Summary

Update the API test workflow so both Dev and UAT jobs resolve the currently deployed uni-region version before checkout. This keeps the test code aligned with the service version running in each environment.

## Changes

- Add `find-dev-deployed-version` to query the Dev `/api/version` endpoint and output the deployed checkout ref.
- Make `API-Tests-Dev` depend on that resolver and checkout the resolved ref before running tests.
- Rename the UAT resolver job to `find-uat-deployed-version` and keep its version API lookup wired to the UAT endpoint.
- Use `nscaledev/quality-tooling/.github/actions/uni-find-staging-constellation@main` for both resolver steps.
- Pass `summary-title` to the resolver action so quality-tooling writes Dev and UAT version summaries directly to `$GITHUB_STEP_SUMMARY`.
- Keep `fallback-to-constellation: false` so a failed deployed-version lookup does not silently test the wrong code.

## Dependency

- Requires nscaledev/quality-tooling#19 to be merged into `main` so the resolver action supports pseudo-version refs and action-owned step summaries.

## Validation

- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/test-api.yaml"); puts "yaml ok"'`
- `git diff --check`


spjmurray (approved):
Let us begin!  Actually, can you trigger this manually from the workflow as dev/staging just to double check it all works first?

PR #575, opened 2026-06-25
Allow server controller event permission

squaremo:
Provider create retries now emit Kubernetes events when it reattempts server creation, and will coalesce retries into the event. Therefore, the service account needs event create and patch. Servers are namespaced resources watched across namespaces, so the controller needs event write access at cluster scope rather than only in the release namespace.

(NB lacking these permissions would have added noise to the logs, but didn't interrupt functioning.)


spjmurray (commented):
I swear there are two events APIs... let me just remind myself.

spjmurray (approved):

PR #578, opened 2026-06-25
Stop region API tests routing through the compute service

spjmurray:
The region API integration suite's default client is built from `API_BASE_URL`, which points at the **compute** service — a *consumer* of region, not the service under test. Several "discovery" reads went through that client, so they were exercising compute's proxy (and silently depending on compute being deployed) instead of testing region. This is what produced the long-running, deterministic "external networks invalid region ID → 404" failure: the request was sent to compute, which doesn't expose that region-only endpoint.

Verified against identity's RBAC roles (`charts/identity/values.yaml`) to decide what is genuinely inaccessible vs merely mis-routed:

- **`region:externalnetworks`** is granted only to `platform-administrator` (global) and `kubernetes-service` (global) — **no** organization-scoped role. An org/end-user principal (which the suite runs as) cannot access it at all, and it is being superseded by the loadbalancer API.
- **`region:regions` / `region:flavors` / `region:images`** are granted to every organization role, so the test principal *can* read them directly on the region service — they were just wired to the wrong base URL.

### Commit 1 — Remove external networks integration tests
Genuinely unreachable in this context; drop the specs plus the now-unused `ListExternalNetworks` test client method and endpoint helper.

### Commit 2 — Route region discovery reads at the region service, not compute
Point `ListRegions`/`ListFlavors`/`ListImages` and the raw discovery/auth requests at the region base URL (`regionClient` / `DoRegionRequest`). `GetRegionDetail` is left as-is (untested; `region:regions/detail` is not granted to org roles).

---

### Commit 3 — Use canonical header key in placement microversion test
Unrelated to the above, but folded in to keep this to a single PR: a golangci-lint dependency bump turned on the `canonicalheader` rule, which fails `make lint` on every PR over the pre-existing `"OpenStack-API-Version"` literal in `placement_test.go`. `http.Header.Get` canonicalizes its key, so `"Openstack-Api-Version"` is behaviour-identical; this just unblocks lint.

### DIFF
--- a/pkg/providers/internal/openstack/placement_test.go
+++ b/pkg/providers/internal/openstack/placement_test.go
@@ -278,7 +278,7 @@ func TestPlacementClientResourceProviderAvailable(t *testing.T) {
 			t.Errorf("required query = %q, want %q", got, want)
 		}
 
-		if got, want := r.Header.Get("OpenStack-API-Version"), "placement 1.18"; got != want {
+		if got, want := r.Header.Get("Openstack-Api-Version"), "placement 1.18"; got != want {
 			t.Errorf("microversion header = %q, want %q", got, want)
 		}
 
--- a/test/api/api_client.go
+++ b/test/api/api_client.go
@@ -370,27 +370,6 @@ func (c *APIClient) DeleteImage(ctx context.Context, orgID, regionID, imageID st
 	}
 }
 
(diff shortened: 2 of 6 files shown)


sravanmedarapu-work (approved):
LGTM

PR #583, opened 2026-06-26
Guard server rebuild behind a durable provisioned latch

spjmurray:
The monitor triggers a bounded delete-and-retry ("rebuild") when it
observes a pre-launch provider server in Healthy/Errored, to ride out
flaky Ironic provisioning. Rebuilding a server that has already booted
would destroy data and foreclose debugging or recovery, so the trigger
must be restricted to servers that have never been successfully
provisioned.

The previous guard keyed off LaunchedAt plus the lifecycle Phase. That
held, but the natural "provisioned" signal — the reconciler-owned
Available condition — is unsuitable: it is re-derived every reconcile
and legitimately flips to Errored/Provisioning on a controller restart
against a flaky provider, exactly when a rebuild would be catastrophic.

Add a write-once status.provisionedAt latch, set by the monitor on the
first Running observation and never cleared, surviving restarts and
provider flakiness. Gate the rebuild on it, keeping LaunchedAt and the
post-launch phases as defence in depth so losing any single status field
cannot re-arm the rebuild path. Existing servers are backfilled on their
next Running poll and covered by the LaunchedAt backstop until then.

De-duplicate the rebuild-eligibility predicate into a single shared
ProviderCreateFailure used by both the provisioner and the controller
watch predicate so the trigger and the action cannot drift.

### DIFF
--- a/charts/region/crds/region.unikorn-cloud.org_servers.yaml
+++ b/charts/region/crds/region.unikorn-cloud.org_servers.yaml
@@ -252,6 +252,19 @@ spec:
                   ProviderCreateRetrying is true while the controller is deleting a failed
                   provider server before making another create attempt.
                 type: boolean
+              provisionedAt:
+                description: |-
+                  ProvisionedAt is a write-once latch recording that the server has booted at
+                  least once. It is set from the same Nova launched_at signal as LaunchedAt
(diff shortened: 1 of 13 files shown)


squaremo (approved):
Looks reasonable, well explained and has tests.

PR #587, opened 2026-06-28
Abort server provider create terminally after retry exhaustion

spjmurray:
The bounded provider-create retry was meant to stop after a small cap (default
3), but the exhausted state ran away: the ProviderCreateFailures counter was
seen climbing into the hundreds across ~19 dev servers at a steady ~10s cadence.
Two independent defects combined to cause it.

1. No terminal path. The generic core reconciler requeues every non-ErrYield
   provisioner error on a fixed 10s interval, so returning a bare error at the
   cap looped forever, re-running Provision (and its provider/identity client
   calls) against a failure that cannot self-heal — starving the workqueue and
   robbing other resources of reconcile time.

2. Increment before the cap check. The counter was advanced before the limit was
   tested, so every re-entry into the failure branch bumped it. Even once the
   loop is stopped, a re-reconcile of an already-exhausted server (notably after
   a controller restart, where the failure predicate still holds) would keep
   advancing the counter past the cap.

Fix both at the single decision site in handleProviderCreateRetry:

- At the cap, return the new core provisioners.Terminal disposition instead of a
  bare error. The reconciler parks the resource (writes the Errored condition,
  no requeue) rather than retrying, so the workqueue goes quiet. The underlying
  "provider server create failed after N attempts" detail rides the
  disposition's operator-facing Why(); the user-visible message is the stable
  reason code. (Requires core with terminal dispositions; go.mod bumped.)

- Test the cap against the prospective attempt and clamp the persisted counter to
  maxAttempts, advancing it only on the retry path. The counter now settles at
  the cap and cannot drift on re-reconcile or restart; an already-drifted
  counter heals back down to the cap on its next pass.

The terminal state is sticky by design: recovery is out of band. Once the
underlying fault is fixed an operator resets ProviderCreateFailures, which
re-arms the retry on the next reconcile. The disposition is Terminal (not
UserActionRequired) precisely because there is no automatic, spec-driven
recovery.

---

**Dependency:** requires `core` >= v1.18.0 for the terminal dispositions (nscaledev/uni-core#306, merged & tagged). `go.mod` is pinned to v1.18.0 — no longer a merge blocker.

**Relation to the counter-guard approach (#586):** complementary, and a superset. #586 stops the counter advancing past the cap; this does that too (clamp, not just guard, so it also heals already-drifted counters and cannot drift across controller restarts), and additionally eliminates the fixed-interval requeue of an exhausted create — the workqueue starvation that a counter-only fix leaves in place.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/spf13/pflag v1.0.10
 	github.com/spjmurray/go-util v0.1.3
 	github.com/stretchr/testify v1.11.1
-	github.com/unikorn-cloud/core v1.17.1
+	github.com/unikorn-cloud/core v1.18.0
 	github.com/unikorn-cloud/identity v1.17.8
 	go.opentelemetry.io/otel v1.43.0
 	go.opentelemetry.io/otel/sdk/metric v1.43.0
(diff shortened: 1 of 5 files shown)


mkjpryor (approved):
LGTM, FWIW

PR #588, opened 2026-06-28
Set all OpenStack compute, network and block storage quotas to unlimited

spjmurray:
Expand the quota override in the OpenStack provisioner so that every quota
field the gophercloud compute (nova), networking (neutron) and block storage
(cinder) clients expose is set to -1 (unlimited), rather than just the subset
we previously overrode.

Quotas are handled globally, not on a per-region basis, so unconditionally
removing all per-project limits in the region provisioner is safe.


mkjpryor (approved):
LGTM

PR #597, opened 2026-07-08
Fail fast on malformed cloud-init at server create

alex-emery:
Part of [INST-1075](https://linear.app/nscale-workspace/issue/INST-1075/fail-fast-on-malformed-cloud-init) (INST-1073 customer issues / release work). Companion: nscaledev/uni-compute#310. Single commit, rebased on main (SSH injection modes).

## Problem

User-data was only validated at the v2 server create boundary when an SSH certificate authority was referenced. Without one, malformed cloud-init passed straight through — failing mid-provision when managed augmentation parses it, or silently inside the guest at boot — leaving customers without an actionable error.

## Change

- **`pkg/userdata` (new)** — owns the API-boundary contract, per review feedback that the *whole* 422 behaviour (status + message) must be shared across services by construction, not convention. `userdata.Validate(userData, managed)` returns the ready `HTTPUnprocessableContent` carrying the parser's specific reason. This repository's v2 server handler and the compute instances handler both call it; it lives outside the handler tree so compute doesn't import another service's handler layer (matching the existing accepted cross-repo surface: `constants`/`ids`/`openapi`/`provisioners/managers/server`).
- **`pkg/provisioners/managers/server/cloudinit.go`** — validation failures now return a typed `*UserDataError{Reason}` that unwraps to `ErrConsistency` (provisioner-side classification and every message byte-identical), so the boundary recovers the caller-facing reason with `errors.As` instead of string surgery. Exports `ValidateUserData` (gzip permitted: passed through unmodified when unmanaged) alongside the existing `ValidateManagedUserData` (gzip rejected: managed augmentation can't merge into it). Multipart detection is now header-order independent (`MIME-Version:` first is valid, as cloud-init accepts), and the unrecognized-format message states what a valid first line looks like.
- **`pkg/handler/server/client_v2.go`** — create validation is unconditional and one line: `userdata.Validate(...)`. `UpdateV2` intentionally still skips it (user-data is only consumed at bootstrap; see existing comment).

## Tests

- `pkg/userdata`: table-driven canonical-422 contract (dispatch, gzip both modes, 422 type, no sentinel leak).
- Parser: every recognized sentinel, multipart edge cases (order-independent, missing boundary, malformed headers, truncated part), CRLF, typed-error contract (`errors.As` + `errors.Is(ErrConsistency)`).
- Handler: no-CA malformed → 422 with actionable message (red-first), valid payloads stored byte-identical, gzip-with-CA → 422.
- Integration (`test/api`): malformed → 422 asserting `error` + `error_description` body fields; valid cloud-config round-trips.

Also verified live on a kind+stgstack dev env through the console: malformed user-data → immediate 422 toast with the actionable message and no CRs created; valid cloud-config and MIME-Version-first multipart → provisioned with user-data byte-identical in both services' CRs.

**Merge ordering:** this must merge and tag before uni-compute#310, which imports `pkg/userdata`.

### DIFF
--- a/pkg/handler/server/client_v2.go
+++ b/pkg/handler/server/client_v2.go
@@ -51,7 +51,7 @@ import (
 	"github.com/unikorn-cloud/region/pkg/openapi"
 	"github.com/unikorn-cloud/region/pkg/providers"
 	"github.com/unikorn-cloud/region/pkg/providers/types"
-	servermanager "github.com/unikorn-cloud/region/pkg/provisioners/managers/server"
+	"github.com/unikorn-cloud/region/pkg/userdata"
 
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
@@ -312,18 +312,6 @@ func generateUserData(in *[]byte) []byte {
 	return *in
 }
 
-func validateUserDataForSSHCertificateAuthority(sshCertificateAuthorityID *string, userData *[]byte) error {
-	if sshCertificateAuthorityID == nil || userData == nil || len(*userData) == 0 {
-		return nil
-	}
-
-	if err := servermanager.ValidateManagedUserData(*userData); err == nil {
-		return nil
-	}
-
-	return errors.HTTPUnprocessableContent("userData must be a recognized cloud-init format when sshCertificateAuthorityId is specified")
-}
-
 func (c *ClientV2) validateSSHCertificateAuthorityReference(ctx context.Context, scope identityids.ProjectScopeReader, sshCertificateAuthorityID *string) error {
 	if sshCertificateAuthorityID == nil {
 		return nil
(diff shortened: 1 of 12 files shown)


pkg/handler/server/client_v2.go
```diff
@@ -293,16 +294,32 @@ func generateUserData(in *[]byte) []byte {
...
 }
 
-func validateUserDataForSSHCertificateAuthority(sshCertificateAuthorityID *string, userData *[]byte) error {
-	if sshCertificateAuthorityID == nil || userData == nil || len(*userData) == 0 {
+// validateUserData rejects user-data that is not recognizable cloud-init, so a
+// malformed payload surfaces as a 422 at create time rather than failing
+// mid-provision (when managed augmentation parses it) or silently inside the
+// guest at boot. When an SSH certificate authority is referenced the payload
+// must also support managed cloud-init augmentation, which excludes gzip.
+func validateUserData(sshCertificateAuthorityID *string, userData *[]byte) error {
```
spjmurray: This smells like repetition waiting to happen in both compute and placements.  Some kind of common handler helper function may work better?

spjmurray: To clarify my original point (I was probably too terse): the concern isn't just factoring out the parse check — it's keeping the *API-level behaviour* identical across handlers. Compute (and placements, if it grows user-data) must return the same 422 with the same shape/message, and right now that contract is only enforced by convention.

What this PR shares is the primitive (`ValidateUserDataInput` + `UserDataValidationReason`), but each handler still re-assembles the HTTP error itself:

```go
return errors.HTTPUnprocessableContent("userData must be a recognized cloud-init format: " + servermanager.UserDataValidationReason(err))
```

So the follow-up will copy that wrapping into compute, and nothing stops the message/status drifting apart later. A single exported handler helper that returns the ready `HTTPUnprocessableContent` (sshCA, userData) → error means every caller gets the same 422 by construction, with nothing to duplicate.

Fine to keep the split for this emergency fix — but can we commit to introducing the single handler-level helper in the compute follow-up so the 422 behaviour is guaranteed consistent rather than convention-based?

spjmurray (approved):
Good enough 👍🏻

PR #598, opened 2026-07-08
Make the monitor the sole owner of the server MAC address

spjmurray:
The MAC address was populated early by the reconciler from the freshly
created Neutron port. That value is correct for VMs but wrong for baremetal:
Ironic rebinds the port to the physical NIC's MAC asynchronously during
provisioning, so the Neutron-generated MAC captured at port-create time is
stale and never corrected once the reconciler goes to sleep.

Move ownership of status.macAddress to the health monitor, the only component
still observing a server after it is provisioned. It records the MAC from the
Nova server response (the port MAC rides inline in addresses, reused from the
poll's existing GetServer call, so there is no extra provider call) once the
server reaches ACTIVE, the barrier at which the port MAC is guaranteed bound
for VMs and baremetal alike. This is one code path for both.

The MAC is only ever written, never cleared: gating on ACTIVE and skipping an
empty read means a transient port-read miss cannot unset a held value, while
an unconditional write of a valid MAC self-heals any drift (the optimistic
status PATCH makes a same-value write a no-op). Accordingly the reconciler's
early port-create writes and the provider-create retry reset no longer touch
the field, and ownership is documented on the CRD and in the package READMEs.

Manually verified read-only against a live production cluster:
- A baremetal compute node reporting Nova ACTIVE had its port bound to a real
  hardware NIC MAC (a physical vendor OUI), not a Neutron-generated fa:16:3e
  address, confirming Ironic completes the rebind by ACTIVE.
- That MAC is absent from `openstack server show` (the CLI collapses addresses
  to bare IPs) but present on the bound port, i.e. in the same Nova addresses
  payload the monitor already reads.
- A throwaway Go probe confirmed gophercloud's servers.Server.Addresses map
  preserves the OS-EXT-IPS-MAC:mac_addr key so the value is reachable in code.

### DIFF
--- a/charts/region/crds/region.unikorn-cloud.org_servers.yaml
+++ b/charts/region/crds/region.unikorn-cloud.org_servers.yaml
@@ -233,8 +233,17 @@ spec:
                 format: date-time
                 type: string
               macAddress:
-                description: MACAddress is the MAC address of the server's primary
-                  network interface.
+                description: |-
+                  MACAddress is the MAC address of the server's primary (single) network
+                  interface. It is owned exclusively by the monitor (the health checker),
+                  not the reconciler: it is recorded from the Nova server response once the
+                  server reaches ACTIVE, the barrier at which the port MAC is guaranteed
+                  bound for VMs and baremetal alike. For baremetal Ironic rebinds the port
+                  to the real NIC MAC asynchronously during deploy, so the value observed
+                  earlier (e.g. at port-create time) is the ephemeral Neutron MAC and is not
+                  trustworthy. It is only ever written, never cleared: the provider-create
+                  retry reset leaves it intact and a transient port-read miss cannot unset
+                  it, while an authoritative ACTIVE read self-heals any drift.
                 type: string
               phase:
                 description: Phase is the current lifecycle phase of the server.
--- a/pkg/providers/internal/openstack/provider.go
+++ b/pkg/providers/internal/openstack/provider.go
@@ -19,6 +19,7 @@ package openstack
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"math/big"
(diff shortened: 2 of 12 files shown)


charts/region/crds/region.unikorn-cloud.org_servers.yaml
```diff
@@ -225,8 +225,17 @@ spec:
                 format: date-time
                 type: string
               macAddress:
-                description: MACAddress is the MAC address of the server's primary
-                  network interface.
+                description: |-
+                  MACAddress is the MAC address of the server's primary (single) network
```
squaremo, line 237: It's slightly odd having this description mentioning the controllers in the CRD spec. It's a private type, so no harm especially (unless/until it becomes a public type).

pkg/providers/internal/openstack/provider.go
```diff
@@ -2153,6 +2154,76 @@ func setServerHealthStatus(server *unikornv1.Server, openstackserver *servers.Se
 	server.StatusConditionWrite(unikornv1core.ConditionHealthy, status, reason, message)
 }
 
+// novaServerAddress is a single entry in a Nova server's `addresses` map. It
+// exists only because gophercloud's servers.Address type omits the port MAC
+// that Nova reports under OS-EXT-IPS-MAC:mac_addr, so we decode the addresses
+// into this to read the MAC with a real type rather than untyped assertions.
+type novaServerAddress struct {
+	//nolint:tagliatelle // Nova API field name, fixed by the OpenStack compute API.
```
squaremo, line 2162: :+1: for the explanation

pkg/providers/internal/openstack/provisioning_status_test.go
```diff
@@ -209,6 +210,59 @@ func TestSetServerPhaseStopped(t *testing.T) {
...
+			},
+		},
+	}}
+	identity := &unikornv1.Identity{}
+	server := &unikornv1.Server{Spec: unikornv1.ServerSpec{
+		FlavorID: "vm",
+		Networks: []unikornv1.ServerNetworkSpec{{ID: "ee2b52e3-a844-42bd-864d-a9ff2f39a026"}},
+	}}
+
+	provider := &Provider{
```
squaremo, line 241: Jeepers. Machine is not afraid of nesting.

squaremo (approved):
Looks sound to me ⭐

PR #602, opened 2026-07-08
Add volume storage model

pcriv:
## Summary

Adds the Region internal `Volume` storage model as non-behavioral plumbing for UNI block storage volumes.

## Changes

- Adds `Volume`, `VolumeList`, `VolumeSpec`, and `VolumeStatus` to the Region CRD model.
- Adds generated Volume CRD and deepcopy artifacts.
- Adds `VolumeClaimRef` with `kind` and `id`, including CRD validation for the supported `Server` claim kind.
- Adds typed `VolumeID`, parse helpers, and test-only must-parse support.
- Keeps Volume on the standard UUID v4 resource identity pattern; no per-network volume name uniqueness key is stored.
- Adds shared CRD schema validation test helpers and Volume claim-reference schema coverage.
- Updates package docs to describe Volume as Region-owned, network-anchored, quota-carrying, and to document the future `Network -> Volume` graph edge.

## Scope Notes

This intentionally does not add public handlers, OpenAPI endpoints, provider calls, controller reconciliation, or extra Region OpenStack VolumeClass configuration. Region VolumeClass configuration is handled separately by STO-119 and is already on `main`.

## Validation

- `make -B generated charts/region/crds`
- `make license`
- `make validate`
- `make generate`
- `make lint`
- `make test-unit`


pkg/apis/unikorn/v1alpha1/types.go
```diff
@@ -744,6 +744,84 @@ type LoadBalancerStatus struct {
...
+	// Tags are an abitrary list of key/value pairs that a client
+	// may populate to store metadata for the resource.
+	Tags unikornv1core.TagList `json:"tags,omitempty"`
+	// NetworkID is the network that anchors the volume's region, identity,
+	// organization, and project scope.
+	NetworkID string `json:"networkID"`
+	// VolumeName is the immutable per-network natural key used to derive the
+	// volume's deterministic resource ID. Mutable display names remain stored
+	// in metadata labels.
+	VolumeName string `json:"volumeName"`
```
nsricardor: Is per-network name uniqueness actually a requirement? Cinder doesn't enforce unique names, and we only ever reference volumes by ID (e.g. ClaimRef.ID), never by name. 
Unless something concretely needs name-based volumes, I'd drop volumeName and follow the v4 pattern  like every sibling resource.

pkg/apis/unikorn/v1alpha1/types.go
```diff
@@ -744,6 +744,84 @@ type LoadBalancerStatus struct {
...
+	VolumeClaimKindServer VolumeClaimKind = "Server"
+)
+
+type VolumeStatus struct {
+	// ObservedGeneration is the most recent generation observed by the controller.
+	ObservedGeneration *int64 `json:"observedGeneration,omitempty"`
+	// Current service state of a volume.
+	Conditions []unikornv1core.Condition `json:"conditions,omitempty"`
+	// ProviderID is the provider-assigned backing volume identifier.
+	ProviderID *string `json:"providerID,omitempty"`
```
nsricardor: I'd drop status.providerID. It goes against our documented convention: controllers re-find provider resources by deterministic name lookup (see Network, SecurityGroup), and pkg/providers/README.md explicitly calls mirrored provider-state records a last resort.

pkg/apis/unikorn/v1alpha1/types.go
pcriv: Addressed in eeb8a90: dropped `VolumeSpec.VolumeName`, removed the deterministic `(networkID, volumeName)` ID helpers/namespace, regenerated the Volume CRD, and updated the docs/tests to state that Volume uses the standard UUID v4 identity pattern.

pkg/constants/constants.go
```diff
@@ -75,6 +75,9 @@ const (
 	// ServerLabel creates an indexable linkage between resources and an
 	// owning entity.
 	ServerLabel = "regions.unikorn-cloud.org/server-id"
+	// VolumeLabel creates an indexable linkage between resources and an
+	// owning block storage volume.
+	VolumeLabel = "regions.unikorn-cloud.org/volume-id"
```
nsricardor: I don't think we need this label. The server/volume relationship is already covered from both ends in the specs (the server references its volume IDs, and the volume points back via spec.claimRef).
Snapshots aren't in v1, so there are no volume-owned dependents that would need it - we can reintroduce it if that changes.
If removed, please also revert the pkg/constants/README.md changes too.

nsricardor (dismissed):
Approving - nice work! The CRD captures the requirements for block storage well, and the model follows the existing patterns in the repo. I left one minor comment about removing constants.VolumeLabel, nothing blocking.
@squaremo tagging you in case you have any concerns before this merges.

squaremo (dismissed):
No concerns, if it's the right shape for your design, it's good :+1:

On git history: I expect this is going to be squashed rather than `git merge --no-ff`. Nonetheless, if I came back here to figure out, say, where a mistake was introduced, the several commits going back and forth on the particulars of the type would make it harder to understand the intent behind the change. In the future, when you change your mind about something mid-branch, use a fixup or squash commit, then `git rebase --autosquash` to straighten out the history. (The machines are pretty good at doing this for you, when you ask.)

nsricardor (dismissed):

nsricardor (approved):

PR #603, opened 2026-07-08
Add region volumeclass configuration

pcriv:
## Summary

Implements Linear STO-119 by adding Region-level OpenStack block storage configuration for VolumeClass inventory.

- Adds `openstack.blockStorage.volumeClasses` to the Region CRD storage model.
- Supports provider volume class ID selection plus user-facing metadata for media, maximum performance caps, and encryption signals.
- Wires the new config through Helm values, chart schema, and Region manifest templating.
- Documents VolumeClass as Region-scoped provider inventory configuration, not a project-owned resource.

## Scope

This intentionally does not add Volume CRD types, VolumeID helpers, resource constants, Volume controllers, provider discovery/lifecycle implementation, API handlers, or OpenAPI endpoints. Those are covered by separate STO-116 child tickets.

## Review follow-up

- Removed `replicated` from VolumeClass metadata until backend-specific storage capabilities are modelled.
- Renamed `iops`/`throughputMiBps` to `maxIOPS`/`maxThroughputMiBps` and documented them as caps, not guarantees.

## Verification

- `make touch`
- `make license`
- `make validate`
- `make lint`
- `make generate`
- `make test-unit`

Replaces #600 after branch rename.


pkg/apis/unikorn/v1alpha1/types.go
```diff
@@ -291,6 +293,65 @@ type RegionOpenstackNetworkSpec struct {
...
+	// Media describes the backing storage medium.
+	Media VolumeClassMedia `json:"media,omitempty"`
+	// Performance describes advertised performance characteristics.
+	Performance *VolumeClassPerformanceSpec `json:"performance,omitempty"`
+	// Encrypted indicates that volumes provisioned from this class are encrypted
+	// at rest by the provider.
+	Encrypted bool `json:"encrypted,omitempty"`
+	// Replicated indicates that volumes provisioned from this class are replicated
+	// by the provider.
+	Replicated bool `json:"replicated,omitempty"`
```
nsricardor: Suggest removing Replicated for now. Volume replication is driver/backend dependent — Cinder itself doesn't replicate blocks generically, so this flag would advertise a property the provider can't uniformly back. Let's wait until the storage strategy (Chad) defines how backend-specific capabilities are modelled before exposing replication in the user-facing inventory.

pkg/apis/unikorn/v1alpha1/types.go
```diff
@@ -291,6 +293,65 @@ type RegionOpenstackNetworkSpec struct {
...
+}
+
+type VolumeClassPerformanceSpec struct {
+	// IOPS describes the advertised input/output operations per second.
+	// +kubebuilder:validation:Minimum=1
+	IOPS *int `json:"iops,omitempty"`
+	// Throughput describes advertised throughput in mebibytes per second.
+	// +kubebuilder:validation:Minimum=1
+	Throughput *int `json:"throughputMiBps,omitempty"`
```
nsricardor: iops/throughputMiBps are ambiguous: providers express performance in different shapes (max, baseline+burst, IOPS proportional to volume size), and Cinder QoS is typically a limit/cap, not a guaranteed reservation. Make it explicit these are maximums - e.g. maxIOPS/maxThroughputMiBps, or nest them under a max/limits block, plus doc comments stating they are caps, not guarantees.

pkg/apis/unikorn/v1alpha1/types.go
```diff
@@ -291,6 +293,64 @@ type RegionOpenstackNetworkSpec struct {
...
+	// ID is the immutable provider identifier for the volume class. For OpenStack,
+	// this is the Cinder volume type ID.
+	ID string `json:"id"`
+	// Media describes the backing storage medium.
+	Media VolumeClassMedia `json:"media,omitempty"`
+	// Performance describes advertised performance characteristics.
+	Performance *VolumeClassPerformanceSpec `json:"performance,omitempty"`
+	// Encrypted indicates that volumes provisioned from this class are encrypted
+	// at rest by the provider.
+	Encrypted bool `json:"encrypted,omitempty"`
```
nsricardor, line 343: Operator-authored metadata is the right call: encryption and QoS are often configured natively on the backend where Cinder can't see them, so we can't reliably derive these fields from Cinder volume type metadata.
Minor suggestion: capture this rationale in the field doc comments or ADMIN.md, like FlavorMetadata does for why OpenStack can't be the source of truth.

nsricardor (dismissed):
Approving - The API shape is sound and it follows the established Flavour pattern. 
@squaremo tagging you in case you have any concerns before this merges.

squaremo (approved):
This looks sensible to me :+1:

PR #604, opened 2026-07-09
Add server volume attachment storage fields

pcriv:
## Summary

Adds the Server-side storage model needed for attach-existing-only block volume attachment work.

## Changes

- Adds `Server.Spec.Volumes` for desired attachment intent referencing existing Region `Volume` resources by ID.
- Adds `Server.Status.Volumes` for per-volume observed attachment rows, including provisioning status, optional guest device name, and message.
- Uses Kubernetes map-list semantics so desired and observed volume rows are keyed by Volume ID.
- Regenerates Server CRD/deepcopy artifacts.
- Updates package docs for the stored Server volume attachment shape.

## Scope Notes

This intentionally stays out of Nova/provider logic, controllers, and public API endpoints. It only introduces the persisted Region storage shape for later controller/API work.

## Validation

- `make touch`
- `make license`
- `make validate`
- `make lint`
- `make generate`
- `make charts/region/crds`
- `go test ./pkg/apis/unikorn/v1alpha1 -count=1`
- `make test-unit`


pkg/apis/unikorn/v1alpha1/types.go
```diff
@@ -1156,6 +1170,23 @@ type ServerStatus struct {
...
+}
+
+type ServerVolumeStatus struct {
+	// ID is the Volume resource ID this attachment row describes.
+	ID string `json:"id"`
+	// ProvisioningStatus indicates whether the volume attachment is ready,
+	// failed, or being provisioned/deprovisioned.
+	ProvisioningStatus AttachmentProvisioningStatus `json:"provisioningStatus"`
+	// Message is a human-readable description of the current attachment state.
+	Message string `json:"message"`
```
nsricardor, line 1192: we need something to express the device name where the volume is attached, something like:
```
     // The guest OS device name where OpenStack presents the attached volume
     Device *string json:"message,omitempty"`
```

pkg/apis/unikorn/v1alpha1/crd_schema_test.go
```diff
@@ -88,7 +90,11 @@ func newCRDValidator(t *testing.T, crdFile string) crdValidator {
 func toUnstructured(t *testing.T, resource any) map[string]any {
 	t.Helper()
 
-	out, err := kruntime.DefaultUnstructuredConverter.ToUnstructured(resource)
+	data, err := json.Marshal(resource)
```
squaremo: Why has this changed? It's not obvious to me.

pcriv: this are leftover of some tests that I added but later removed, I'll remove it too

squaremo (approved):

PR #607, opened 2026-07-10
Add VolumeClass contract and OpenStack discovery

pcriv:
## Summary

- Add provider-neutral `VolumeClass` inventory types and expose them through the common Region provider discovery contract. This covers STO-121.
- Implement OpenStack Cinder volume type discovery with public visibility filtering, Region selector filtering, metadata enrichment from STO-119 configuration, and provider-neutral conversion. This covers STO-122.
- Cache OpenStack block-storage inventory through the refreshable service-client state so repeated `VolumeClasses` calls reuse the Cinder volume-type cache while still refreshing on Region configuration or credential changes.
- Add Kubernetes and simulated provider implementations, regenerated provider mocks, provider docs, focused OpenStack tests, and the provider-contract OpenStack catalog fixture update.

## Scope

This intentionally stays provider-facing only. It does not add Volume CRDs, Volume lifecycle methods, server attach/detach contracts, controllers, handlers, storage-model integration, or public OpenAPI endpoints.

## Validation

- `make touch`
- `make license`
- `make validate`
- `make lint`
- `make generate`
- `git status --porcelain`
- `make test-unit`

Closes STO-121.
Closes STO-122.

### DIFF
--- a/pkg/providers/internal/openstack/blockstorage.go
+++ b/pkg/providers/internal/openstack/blockstorage.go
@@ -149,7 +149,7 @@ func (c *BlockStorageClient) UpdateQuotas(ctx context.Context, projectID string)
 }
 
 func (p *Provider) VolumeClasses(ctx context.Context) (types.VolumeClassList, error) {
-	blockStorage, err := p.blockStorage(ctx)
+	blockStorage, err := p.openstack.blockStorage(ctx)
 	if err != nil {
 		return nil, err
 	}
@@ -162,17 +162,6 @@ func (p *Provider) VolumeClasses(ctx context.Context) (types.VolumeClassList, er
 	region, _ := p.openstack.regionSnapshot()
 
 	return convertVolumeClasses(region, resources), nil
-}
-
-func (p *Provider) blockStorage(ctx context.Context) (VolumeTypeInterface, error) {
-	region, credentials, err := p.openstack.regionRefresh(ctx)
-	if err != nil {
-		return nil, err
-	}
-
-	providerClient := NewPasswordProvider(region.Spec.Openstack.Endpoint, credentials.userID, credentials.password, credentials.projectID)
-
-	return NewBlockStorageClient(ctx, providerClient, region.Spec.Openstack.BlockStorage)
 }
 
 func convertVolumeClasses(region *unikornv1.Region, resources []volumetypes.VolumeType) types.VolumeClassList {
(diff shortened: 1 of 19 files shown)


pkg/providers/types/types.go
```diff
@@ -88,6 +88,42 @@ type GPU struct {
...
+)
+
+// VolumeClass represents provider block-storage inventory exposed by a Region.
+type VolumeClass struct {
+	// ID is the immutable provider identifier.
+	ID string
+	// Name is the provider display name.
+	Name string
+	// Description is the provider display description.
+	Description string
```
nsricardor, line 107: where name and description is coming from?

pcriv: They are part of the response from Openstack, should we ignore them?

pkg/providers/internal/openstack/blockstorage.go
```diff
@@ -99,3 +147,105 @@ func (c *BlockStorageClient) UpdateQuotas(ctx context.Context, projectID string)
...
+
+func (p *Provider) blockStorage(ctx context.Context) (VolumeTypeInterface, error) {
+	region, credentials, err := p.openstack.regionRefresh(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	providerClient := NewPasswordProvider(region.Spec.Openstack.Endpoint, credentials.userID, credentials.password, credentials.projectID)
+
+	return NewBlockStorageClient(ctx, providerClient, region.Spec.Openstack.BlockStorage)
```
nsricardor: The cache never takes effect in production. blockStorage() builds a new PasswordProvider + BlockStorageClient on every call, so each VolumeClasses request pays a full Keystone auth plus a Cinder list against a fresh, empty one-hour cache. Every other service client (identity/compute/image/network) is built once in newServiceClientState and rebuilt only on config/secret change — the block storage client should live there too, with an accessor like openStackClients.compute().

The unit test doesn't catch this because it constructs one client directly and calls it twice, which production never does. Once the client is long-lived, please add a provider-level test (via NewTestProvider) calling Provider.VolumeClasses twice and asserting a single Cinder request, plus coverage for the default branch: nil/empty selector returns all public volume types.

nsricardor (approved):
Approved ✅

I confirmed that the implementation for BlockStorageClient.GetVolumeTypes follows the same pattern as ComputeClient.GetFlavors (1-hour cache, admin listing with private types filtered out, and selector filtering with the same nil/empty behavior).
Provider.VolumeClasses also follows the same flow as Provider.Flavors: fetch → snapshot → metadata enrichment.
This doesn't introduce any backward compatibility issues, and since it's not publicly exposed yet, it's safe to go.

PR #620, opened 2026-07-23
Make OpenStack volume class selection fail closed

pcriv:
## Summary

- Return only public Cinder volume types whose provider IDs are explicitly listed in `spec.openstack.blockStorage.volumeClasses.selector.ids`.
- Return an empty VolumeClass inventory when the VolumeClass configuration, selector, or selector IDs are missing or empty.
- Document the fail-closed behavior and rollout prerequisite, and regenerate the Region CRD descriptions.

## Why

OpenStack VolumeClass discovery previously failed open when selector configuration was absent or empty. That could expose newly discovered provider volume types without an operator explicitly curating them.

This change makes the Region selector a strict allowlist so inventory exposure always requires operator intent.

## Impact

Curated provider IDs must be added to affected production Region CRs before rollout, otherwise their VolumeClass inventory will become empty.

There are no OpenAPI, Helm template, structural schema, Volume lifecycle, Volume CRUD, or server attachment changes.

## Validation

- `make touch`
- `make license`
- `make validate`
- `make lint`
- `make generate`
- `make test-unit`
- Focused OpenStack VolumeClass regression tests
- Standards and STO-150 scope reviews: no findings

Linear: [STO-150](https://linear.app/nscale-workspace/issue/STO-150/uni-region-make-openstack-volumeclass-selector-fail-closed)

### DIFF
--- a/charts/region/crds/region.unikorn-cloud.org_regions.yaml
+++ b/charts/region/crds/region.unikorn-cloud.org_regions.yaml
@@ -197,8 +197,9 @@ spec:
                       volumeClasses:
                         description: |-
                           VolumeClasses defines which provider volume classes are eligible for Region
-                          inventory export and what user-facing metadata is attached to them. If not
-                          defined, then all provider volume classes are eligible.
+                          inventory export and what user-facing metadata is attached to them. Provider
+                          volume classes are excluded unless their IDs are explicitly selected. If this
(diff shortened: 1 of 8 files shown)


nsricardor (approved):
Nice, focused change - exactly the right call before the first release.

PR #618, opened 2026-07-22
Publish VolumeClass OpenAPI contract

pcriv:
## Summary

- publish the public `GET /api/v2/volumeclasses` Region API contract
- support optional, repeatable `regionID` filtering following the file-storage-class inventory pattern
- define stable resource metadata and provider-neutral media, encryption, and advertised-performance fields
- regenerate OpenAPI types, client, router, and embedded schema artifacts
- keep server compilation intact with an endpoint-specific canonical `500` scaffold until provider-backed list handling is implemented separately

## Scope

This is a wire-contract change only. It does not add Volume lifecycle or CRUD APIs, attachments, controllers, provider discovery, provider implementations, or integration tests, and it has no STO-122 dependency.

This PR is stacked on #607, which introduces the provider-neutral VolumeClass domain contract.

## Verification

- `make touch`
- `make license`
- `make validate`
- `make lint`
- `make generate`
- `make test-unit`
- focused OpenAPI, handler, and server compile tests
- independent standards and ticket-scope reviews

## Known repository issue

`make validate-docs` is currently blocked by the unchanged target referencing the removed `github.com/unikorn-cloud/core/hack/docs` package. The available `hack/docs-generate` replacement does not support the expected `--dry-run` option.

Linear: [STO-145](https://linear.app/nscale-workspace/issue/STO-145/uni-region-publish-volumeclass-openapi-contract)

PR #621, opened 2026-07-24
Implement VolumeClass API handler

pcriv:
## Summary

- implement the read-only v2 VolumeClass inventory handler over the provider-neutral discovery interface
- list only Regions visible to the caller and treat repeated `regionID` values as selectors, omitting missing or inaccessible Regions
- map provider-neutral metadata, media, performance, and encryption fields without exposing provider-specific details
- return deterministic results ordered by Region, class name, and class ID
- return empty lists for empty Region configuration, unmatched filters, and empty provider inventories while failing the request on provider errors
- update the OpenAPI contract, package documentation, generated clients, and unit coverage

## Why

Callers need a provider-independent way to discover block-storage capabilities before creating storage resources. Keeping filtering aligned with other v2 list endpoints allows clients to request several Regions and receive the visible subset without turning the operation into a sequence of resource lookups.

## Stack

Depends on #618 (`feat/STO-145-publish-volumeclass-openapi-contract`).

Linear: https://linear.app/nscale-workspace/issue/STO-146/uni-region-implement-volumeclass-api-handler

## Validation

- `make touch`
- `make license`
- `make validate`
- `make lint`
- `make generate`
- clean generated worktree check
- `make test-unit`
- `go build ./...`

### DIFF
--- a/pkg/handler/conversion/volumeclass.go
+++ b/pkg/handler/conversion/volumeclass.go
@@ -1,0 +1,66 @@
+/*
+Copyright 2026 Nscale.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package conversion
+
+import (
+	coreapi "github.com/unikorn-cloud/core/pkg/openapi"
+	regionids "github.com/unikorn-cloud/region/pkg/ids"
+	"github.com/unikorn-cloud/region/pkg/openapi"
+	"github.com/unikorn-cloud/region/pkg/providers/types"
+
+	"k8s.io/utils/ptr"
+)
+
+func ConvertVolumeClasses(regionID regionids.RegionID, in types.VolumeClassList) openapi.VolumeClassListV2Read {
+	out := make(openapi.VolumeClassListV2Read, len(in))
+
+	for i := range in {
+		out[i] = *convertVolumeClass(regionID, &in[i])
+	}
+
+	return out
+}
+
+func convertVolumeClass(regionID regionids.RegionID, in *types.VolumeClass) *openapi.VolumeClassV2Read {
+	out := &openapi.VolumeClassV2Read{
+		Metadata: coreapi.StaticResourceMetadata{
+			Id:   in.ID,
+			Name: in.Name,
+		},
+		Spec: openapi.VolumeClassV2Spec{
+			RegionId:  regionID,
+			Encrypted: in.Encrypted,
+		},
+	}
+
+	if in.Description != "" {
+		out.Metadata.Description = ptr.To(in.Description)
+	}
+
+	if in.Media != "" {
+		out.Spec.Media = ptr.To(openapi.VolumeClassV2Media(in.Media))
+	}
+
+	if in.Performance != nil {
+		out.Spec.Performance = &openapi.VolumeClassV2Performance{
+			MaxIOPS:            in.Performance.MaxIOPS,
+			MaxThroughputMiBps: in.Performance.MaxThroughput,
+		}
+	}
+
+	return out
+}
--- a/pkg/handler/handler_v2_volumeclass_test.go
+++ b/pkg/handler/handler_v2_volumeclass_test.go
@@ -1,0 +1,490 @@
+/*
+Copyright 2026 Nscale.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+//nolint:testpackage
+package handler
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+
+	coreapi "github.com/unikorn-cloud/core/pkg/openapi"
+	identityapi "github.com/unikorn-cloud/identity/pkg/openapi"
+	"github.com/unikorn-cloud/identity/pkg/rbac"
+	regionv1 "github.com/unikorn-cloud/region/pkg/apis/unikorn/v1alpha1"
+	"github.com/unikorn-cloud/region/pkg/handler/common"
+	"github.com/unikorn-cloud/region/pkg/openapi"
+	mockproviders "github.com/unikorn-cloud/region/pkg/providers/mock"
+	"github.com/unikorn-cloud/region/pkg/providers/types"
+	mockprovider "github.com/unikorn-cloud/region/pkg/providers/types/mock"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// errVolumeClassProvider is returned by provider mocks to exercise error handling.
+var errVolumeClassProvider = errors.New("volume class provider failed")
+
+const (
+	volumeClassTestNamespace       = "volume-class-test"
+	volumeClassReadEndpoint        = "region:volumeclasses:v2"
+	volumeClassTestOrganizationID  = "10101010-1010-4010-a010-101010101010"
+	volumeClassOtherOrganizationID = "20202020-2020-4020-a020-202020202020"
+)
+
+func volumeClassReadContext(ctx context.Context) context.Context {
+	return newOrganisationACLBuilder(volumeClassTestOrganizationID).
+		addEndpoint(volumeClassReadEndpoint, identityapi.Read).
+		buildContext(ctx)
+}
+
+type volumeClassV2TestFixture struct {
+	t         *testing.T
+	ctrl      *gomock.Controller
+	providers *mockproviders.MockProviders
+	handler   *Handler
+}
+
+func newVolumeClassV2TestFixture(t *testing.T, regions ...*regionv1.Region) *volumeClassV2TestFixture {
+	t.Helper()
+
+	objects := make([]client.Object, len(regions))
+	for i := range regions {
+		objects[i] = regions[i]
+	}
+
+	ctrl := gomock.NewController(t)
+	providers := mockproviders.NewMockProviders(ctrl)
+
+	return &volumeClassV2TestFixture{
+		t:         t,
+		ctrl:      ctrl,
+		providers: providers,
+		handler: &Handler{
+			ClientArgs: common.ClientArgs{
+				Client:    fakeClientWithSchema(t, objects...),
+				Namespace: volumeClassTestNamespace,
+				Providers: providers,
+			},
+		},
+	}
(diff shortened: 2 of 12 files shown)


pkg/handler/region/region.go
```diff
@@ -171,6 +171,102 @@ func (c *Client) ListFlavors(ctx context.Context, organizationID identityids.Org
 	return conversion.ConvertFlavors(result), nil
 }
 
+func (c *Client) ListVolumeClasses(ctx context.Context, params openapi.GetApiV2VolumeclassesParams) (openapi.VolumeClassListV2Read, error) {
+	regionIDs, err := c.volumeClassRegionIDs(ctx, params.RegionID)
```
pcriv: **[P1] Enforce endpoint RBAC before provider discovery.** `volumeClassRegionIDs` only applies the Region security allowlist via `CheckAccess`; it does not verify an endpoint permission. For an unrestricted Region, `checkAccess` returns `nil` before inspecting the ACL, and for a restricted Region it accepts organization membership without checking a read operation. The middleware populates the ACL but does not enforce resource permissions, so an authenticated principal with no VolumeClass/Region catalogue read grant can still reach `provider.VolumeClasses`. Please enforce the intended read permission for every selected Region before lookup (per Platform Specification §§4.3 and 7.2) and add a negative test where the ACL contains the organization but lacks that endpoint permission.

pkg/handler/conversion/volumeclass.go
```diff
@@ -0,0 +1,69 @@
...
+
+	return out
+}
+
+func convertVolumeClass(regionID regionids.RegionID, in *types.VolumeClass) *openapi.VolumeClassV2Read {
+	out := &openapi.VolumeClassV2Read{
+		Metadata: coreapi.StaticResourceMetadata{
+			Id:           in.ID,
+			Name:         in.Name,
+			CreationTime: time.Unix(0, 0).UTC(),
```
pcriv: **[P2] Do not publish a fabricated creation timestamp.** Core's `staticResourceMetadata.creationTime` is required and defined as “the time the resource was created,” but `types.VolumeClass` carries no such value. Hard-coding the Unix epoch makes every class claim it was created in 1970 and the new test locks that false data into the API contract. Either carry a real timestamp from the provider, if one exists, or revise the VolumeClass schema to use metadata that does not require `creationTime`.

pcriv (commented):
Found two actionable issues: endpoint RBAC is not enforced before provider discovery, and the response fabricates resource creation timestamps.

pkg/handler/region/region.go
```diff
@@ -171,6 +173,146 @@ func (c *Client) ListFlavors(ctx context.Context, organizationID identityids.Org
...
+	return result, nil
+}
+
+func checkVolumeClassAccess(ctx context.Context, resource *unikornv1.Region) error {
+	if rbac.AllowGlobalScope(ctx, volumeClassReadEndpoint, identityapi.Read) == nil {
+		return nil
+	}
+
+	organizationIDs := rbac.OrganizationIDs(ctx)
+	if resource.Spec.Security != nil && len(resource.Spec.Security.Organizations) > 0 {
```
pcriv: **`organizations: []` edge case diverges from region visibility.**

`checkVolumeClassAccess` treats `len(Security.Organizations) == 0` as unrestricted, while the existing `checkAccess` (line 67) treats only `Organizations == nil` as unrestricted — a non-nil empty list denies everyone. So a Region configured with `security.organizations: []` is hidden from `GET /regions`, yet its volume classes are served to anyone holding the endpoint grant.

This code actually matches the CRD comment ("if not empty limits access") better than `checkAccess` does, but the two paths should agree on the edge case — whichever way, one of them should change or the divergence deserves a caveat note.

pkg/handler/region/region.go
```diff
@@ -49,6 +49,8 @@ var (
 	ErrRegionNotFound = goerrors.New("region doesn't exist")
 )
 
+const volumeClassReadEndpoint = "region:volumeclasses:v2"
```
pcriv: **Rollout dependency**: `region:volumeclasses:v2` must be granted by roles defined in the identity service. Until those roles exist, every non-global-scope caller gets 404 (explicit filter) or an empty list (unfiltered) with no signal that it's a permissions problem. Worth confirming the uni-identity change is queued and sequencing the rollout — the PR description doesn't mention it.

pkg/handler/region/region.go
```diff
@@ -171,6 +173,146 @@ func (c *Client) ListFlavors(ctx context.Context, organizationID identityids.Org
...
+
+		if rbac.AllowOrganizationScopeID(ctx, volumeClassReadEndpoint, identityapi.Read, organizationID) == nil {
+			return nil
+		}
+	}
+
+	return errors.HTTPNotFound()
+}
+
+func (c *Client) checkVolumeClassAccess(ctx context.Context, regionID regionids.RegionID) error {
```
pcriv: **Naming/duplication nit**: package-level `checkVolumeClassAccess(ctx, resource)` and this method share the same name — legal, but the last line reads like recursion at a glance. The existing pattern disambiguates by case (`checkAccess` func vs `CheckAccess` method); suggest e.g. `checkVolumeClassAccessForRegion` for one of them.

The method also duplicates `CheckAccess`'s get-region-or-404 preamble; a shared fetch helper would remove the copy.

pkg/handler/region/README.md
```diff
@@ -33,6 +34,12 @@ provider capability discovery into user-visible region catalogue data.
 - Provider-derived capability reads must resolve the correct provider for the
   selected region rather than trusting client-supplied assumptions.
 - Flavor ordering is intentionally stable and user-facing.
+- VolumeClass inventory is ordered by Region, class name, and class ID. Empty
+  provider inventory remains a non-nil empty API list.
+- Explicit VolumeClass `regionID` filters are deduplicated and checked for both
```
pcriv: Slight overstatement: "checked for both Region visibility and `region:volumeclasses:v2/read` permission" — the code checks only the volumeclass grant against the Region's allowlist; it never invokes the region-visibility `checkAccess`. The outcomes coincide except for the `organizations: []` edge case (see comment on region.go:219) and global-scope holders, so either tighten the wording or align the code.

pkg/handler/handler_v2_volumeclass_test.go
```diff
@@ -0,0 +1,585 @@
...
+	params := openapi.GetApiV2VolumeclassesParams{
+		RegionID: ptr.To(openapi.RegionIDQueryParameter{regionID}),
+	}
+
+	handler.GetApiV2Volumeclasses(response, request, params)
+
+	require.Equal(t, http.StatusNotFound, response.Code)
+}
+
+func TestVolumeClassV2RejectsRegionsWithoutEndpointPermission(t *testing.T) {
```
pcriv: The no-endpoint-grant case is tested only for the explicit-filter path (404). The unfiltered path's behavior — 200 with an empty list when the caller lacks the grant everywhere — is a deliberate semantic worth pinning with a test, since it's the "silent" failure mode of a missing identity role grant.

pcriv (commented):
Re-review of head `b9d7acf` ("Address VolumeClass review feedback").

### What changed
1. **`creationTime` removed from the wire contract** via a dedicated `volumeClassV2Metadata` schema (id/name/description only), with a contract test pinning its absence. Cleaner than any sentinel timestamp — no fabricated lifecycle data, and the rationale is documented in `conversion/README.md` and `openapi/README.md`. ✅
2. **Dedicated RBAC gate added**: `region:volumeclasses:v2/read`. Verified against identity v1.17.8 that `rbac.AllowOrganizationScopeID` checks the grant against the specific organization's ACL entry, so iterating the Region's allowlist correctly computes "caller has the grant in an allowed org". Endpoint naming follows the `region:filestorage:v2` convention. ✅

### Status of previous findings
- CreationTime inconsistency — **resolved** (contract-level fix).
- Sequential fan-out / fail-whole-request — unchanged; documented as intentional, acceptable.
- Redundant double sorting (`sortRegionIDs` vs the final region-aware `SortStableFunc`) — unchanged; minor.
- Media direct cast vs defensive switch — unchanged; minor.
- Missing 400 test for malformed `regionID`; cross-region ordering untested — unchanged; minor.

### New findings
See inline comments. Summary: nothing blocking. The empty-allowlist divergence between `checkAccess` and `checkVolumeClassAccess` is the one I'd resolve before merge; the identity-role rollout dependency is a coordination check; the rest is polish.

Build and all VolumeClass unit tests pass locally on this head.

nsricardor (commented):
Just a few small findings we should address before this is merged

The main concern is that endpoint-level RBAC is no longer enforced at the handler. The request enters discovery immediately, and region:volumeclasses:v2/read is only evaluated later on a per-Region basis. As a result, an authenticated caller without the endpoint permission receives 200 [] for an unfiltered request instead of being rejected. The endpoint authorization check should happen in the handler before invoking the Region client, while the Region client should continue to enforce per-resource visibility independently.

The second issue is that regionID has effectively become a fail-fast lookup rather than a list filter. The current implementation resolves and authorizes every supplied Region up front and returns 404 as soon as one is missing or inaccessible, which changes the semantics from a filtered list operation into a multi-resource GET. This also bakes the behavior into the OpenAPI spec. Existing v2 list endpoints treat regionID as a selector: apply the filter, perform the list, omit inaccessible resources, and naturally return an empty (or partial) result set with 200 when nothing matches. This endpoint should follow the same convention for consistency.

Minor point: the new test file contains a significant amount of duplicated setup and request/response plumbing. It would be worth extracting a small test harness and using table-driven cases to improve maintainability, although this is secondary to the behavioral issues above.

pkg/handler/region/region.go
```diff
@@ -171,6 +180,69 @@ func (c *Client) ListFlavors(ctx context.Context, organizationID identityids.Org
...
+		for _, value := range organizationIDs {
+			organizationID, err := identityids.ParseOrganizationID(value)
+			if err != nil {
+				continue
+			}
+
+			if rbac.AllowOrganizationScopeID(ctx, volumeClassReadEndpoint, identityapi.Read, organizationID) == nil {
+				continue
+			}
+		}
```
nsricardor: The current logic doesn't enforce the intended authorization check. The requirement is that at least one organization must pass rbac.AllowOrganizationScopeID before we include the region's volume classes. Right now, the result of the RBAC check isn't used to gate the rest of the logic, so we always proceed to fetch the volume classes.

From a performance perspective, we should also avoid calling c.Providers.LookupCommon (and subsequently provider.VolumeClasses) unless an organization has already passed the RBAC check. That way we skip the provider lookup and API call entirely for unauthorized requests.

pkg/handler/region/region.go
```diff
@@ -171,6 +180,69 @@ func (c *Client) ListFlavors(ctx context.Context, organizationID identityids.Org
...
+
+func (c *Client) ListVolumeClasses(ctx context.Context, params openapi.GetApiV2VolumeclassesParams) (openapi.VolumeClassListV2Read, error) {
+	regions := &unikornv1.RegionList{}
+
+	if err := c.Client.List(ctx, regions, &client.ListOptions{Namespace: c.Namespace}); err != nil {
+		return nil, err
+	}
+
+	// Filter regions based on RBAC
+	FilterRegions(ctx, regions)
```
nsricardor: This block can be extracted and shared with List(), for example:

```
// listRegions returns the regions in the namespace, filtered by the caller's RBAC.
func (c *Client) listRegions(ctx context.Context) (*unikornv1.RegionList, error) {
      regions := &unikornv1.RegionList{}

      if err := c.Client.List(ctx, regions, &client.ListOptions{Namespace: c.Namespace}); err != nil {
              return nil, err
      }

      FilterRegions(ctx, regions)

      return regions, nil
}
```

List() becomes regions, err := c.listRegions(ctx); ... return convertList(regions), nil, and ListVolumeClasses drops its 6 lines to one

pkg/handler/region/region.go
pcriv: Fixed in https://github.com/nscaledev/uni-region/pull/621/changes/4fbb3a551b28e95339eb7362cddc07dc82e573a9

pkg/handler/region/region.go
pcriv: Fixed in https://github.com/nscaledev/uni-region/pull/621/changes/088811349884278080de820887775d2c45e04135

nsricardor (approved):

PR #627, opened 2026-07-29
Add provider-neutral Volume lifecycle contracts

pcriv:
## Summary

- add a standalone provider capability for Volume create, find, observe, and idempotent delete behavior
- preserve native `unikornv1.Volume` lifecycle intent while returning provider-neutral observed size and status
- regenerate the provider mock and add a compile-time contract regression test
- document the mixed provider boundary and deliberately exclude VolumeClass inventory and server attach/detach operations

## Why

STO-139 establishes the provider boundary needed by the later OpenStack Volume primitives, controller, and monitor work without leaking Cinder resource types into provider consumers.

The Volume capability remains separate from the composite `types.Provider` until concrete provider lifecycle implementations are introduced.

## Review focus

- confirm whether `FindVolume` should return `(bool, error)` or use the repository's `ErrResourceNotFound` convention
- confirm the provider-neutral Volume status vocabulary before the OpenStack mapping is implemented

## Validation

- `make touch`
- `make license`
- `make validate`
- `make lint`
- `make generate`
- generated-code and working-tree audit
- `make test-unit`
- `git diff --check`

Linear: STO-139

### DIFF
--- a/pkg/README.md
+++ b/pkg/README.md
@@ -20,8 +20,10 @@ The useful way to read it is not as a directory tree, but as one system:
 - monitors project observed provider truth back into status and metrics
 - providers bind the region model to real or simulated clouds
 - `Volume` is an internal Region storage model today: a network-anchored,
-  quota-carrying block storage resource whose public API, controller, and
-  provider behavior are deliberately introduced by later tickets
+  quota-carrying block storage resource. Its provider-facing create, find,
+  observe, and delete contract is defined, while concrete provider behavior,
+  public API, controller, and monitor integration are introduced by later
+  tickets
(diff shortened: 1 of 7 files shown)


nsricardor (commented):
I like keeping this work in small PRs, but this contract-only change is difficult to validate because nothing implements or consumes it yet. Could we split it into lifecycle slices instead: first create/delete with a concrete provider implementation and tests, then find/observe plus status mapping, then controller/monitor integration? That would let each PR establish and exercise a real part of the contract, rather than freezing the full interface before its callers and implementations have informed the design.

PR #628, opened 2026-07-30
Implement OpenStack Volume create and delete

pcriv:
## Summary

- add a focused provider-facing Volume capability for create and delete using the native `unikornv1.Volume` lifecycle intent
- implement project-scoped Cinder volume rediscovery, creation, metadata linkage, and idempotent deletion
- use the stable `volume-<Region Volume UUID>` name and exact-match filtered Cinder results
- regenerate provider mocks and update the affected provider package documentation

## Scope

This is the first implementation-backed Volume lifecycle slice from STO-139. Provider observation/status mapping, VolumeClass inventory changes, and server attach/detach operations remain outside this PR.

## Validation

- `make touch`
- `make license`
- `make validate`
- `make lint`
- `make generate`
- generated-code status audit
- `make test-unit`
- `git diff --check`

Linear: STO-139

### DIFF
--- a/pkg/providers/internal/openstack/provider.go
+++ b/pkg/providers/internal/openstack/provider.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"maps"
 	"math/big"
 	"net"
 	"net/http"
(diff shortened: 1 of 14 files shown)


pkg/providers/internal/openstack/provider.go
```diff
@@ -140,7 +142,10 @@ type Provider struct {
 	imageCache *cache.RefreshAheadCache[types.Image, *types.Image]
 }
 
-var _ types.Provider = &Provider{}
+var (
+	_ types.Provider = &Provider{}
+	_ types.Volume   = &Provider{}
```
nsricardor, line 147: Minor: when types.Volume is added to types.Provider, this assertion will be redundant. Please remove it then so we don’t forget later.

nsricardor (approved):

PR #635, opened 2026-08-04
Add configurable OpenStack default architecture

mkjpryor:
## What changed

- Add optional Region configuration at `spec.openstack.defaultArchitecture`, defaulted to `x86_64` and constrained to `x86_64` or `aarch64`.
- Apply the regional fallback to OpenStack flavor and image inventory while preserving explicit per-flavor and Glance metadata precedence.
- Pass the fallback through background image refresh, direct image creation, and server snapshot cache seeding.
- Expose the setting through Helm values, schema validation, and Region rendering.
- Regenerate the Region CRD and deepcopy implementation, and document the precedence rules.

## Why

Some OpenStack regions primarily expose `aarch64` inventory but do not consistently attach architecture metadata to every Nova flavor or Glance image. A Region-scoped fallback keeps inventory resolution inside the provider boundary without introducing a process-wide flag, while preserving existing `x86_64` behavior for omitted configuration.

## Impact

Existing Region objects require no migration and retain their current `x86_64` fallback. Operators can opt an OpenStack region into `aarch64`; explicit flavor or image metadata still wins.

## Validation

- `make license`
- `make validate`
- `make lint`
- `make generate`
- `make test-unit`
- Helm rendering verified for `x86_64` and `aarch64`
- Helm values schema verified to reject `amd64`
- Forced second CRD/deepcopy generation produced an identical content hash

### DIFF
--- a/charts/region/crds/region.unikorn-cloud.org_regions.yaml
+++ b/charts/region/crds/region.unikorn-cloud.org_regions.yaml
@@ -435,6 +435,15 @@ spec:
                           scheduling cluster nodes.  Defaults to "soft-anti-affinity".
                         type: string
                     type: object
+                  defaultArchitecture:
+                    default: x86_64
+                    description: |-
+                      DefaultArchitecture is the fallback architecture for flavors and images
+                      that do not define one explicitly.
+                    enum:
+                    - x86_64
+                    - aarch64
+                    type: string
                   endpoint:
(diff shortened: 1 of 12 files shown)


squaremo (approved):
Seems reasonable to me. Is there any risk involved with rolling this out? If the region is predominantly aarch64 with some unspecified x86_64 hosts or images, could this change stop those hosts/images working where they worked before (and nothing else did ..)?

PR #71, opened 2024-11-06
Add Server Controller part 1

nsricardor:
This PR introduces the Server CRD and its corresponding OpenstackServer CRD within the region service to support the creation and management of workload pools for Slonk through the compute service. 
The PR includes the initial controller implementation, a Dockerfile for building the controller image, and updates to the Helm chart to deploy the server controller. 

This PR focuses on defining the data structures provided by the CRDs and does not include reconciliation logic or the OpenStack provider implementation, allowing for a more manageable PR size.

PR #636, opened 2026-08-04
Rebuild refactor

alex-emery:
## What

Replaces the `Status.Rebuild` state machine with a reader/writer split: the reconciler decides rebuilds from a fresh provider read, and the monitor records what it saw into a new `status.observed` region.

**Breaking**: `Status.Rebuild` (`TargetImageID` + the `Initiated/Rebuilding/Succeeded/Failed` enum) is removed from the Server CRD. A parked `Failed` rebuild no longer exists as a state — a failed rebuild leaves the server `ERROR` on the desired image, surfaced on the monitor's lifecycle axis, and recovery remains another image or a replacement server.

## Why

The marker had two writers holding different models of one field, and every interaction between them needed an ordering argument (forward-only rank rules, two-phase parks, settlement wakes, single-patch liveness invariants). Measurement replaced the assumptions those defenses existed for: Nova commits `image_ref` and `task_state` together under a compare-and-swap in its API layer, before the 202, above the driver. Given that, a fresh read is sufficient to decide every pass, and the marker's write-ahead intent recorded nothing a fresh read cannot reconstruct.

## How

- `reconcileServerImage` is now a seven-row table decided entirely from the fresh `GetServer` (see `pkg/providers/internal/openstack/README.md`). The one destructive row submits only on `ref != desired ∧ launched ∧ quiescent`; an accepted rebuild yields, so `provisioning` falls out of core's yield mapping and the v2 handler's status override is deleted.
- `status.observed` carries `generation` (freshness stamp), `image` (unreadable refs preserve the previous value), and `error` (live state, cleared on an authoritative non-error read; fault code is a bounds-checked pointer so "no fault detail visible" is distinguishable from code 0). An observation never authorizes an action against the provider.
- The server manager wakes the reconciler on any semantic change to the observed region, replacing the `RebuildSettled` level wake. Liveness no longer depends on the monitor: `ErrYield` requeues on a timer.
- `GetServer` re-reads an errored server by ID because Nova (≤2025.2) can omit `fault` from list responses; healthy servers pay no second call.

## Testing

- Unit coverage for the pass table, the observed projection, the wake predicate, and the fault re-read.
- Integration-lane server coverage via the simulated provider is split out to #642 (stacked on this PR). The per-PR rebuild-contract test (real provider against a fake Nova) is follow-up work on the ironic-fake-control/fake-dc rail; until then the contract is guarded by the unit pass-table and nightly e2e.
- **Live end-to-end runs (2026-08-07)**: the region service built from this branch, deployed via the dev-env into kind, drove real rebuilds on the kolla lab through the full stack (API → reconciler → Nova → libvirt/Ironic → monitor → API). The e2e rebuild suite passed on both the QEMU flavor and the Ironic (sushy virtual-metal) flavor: in-place rebuild retaining identity and addresses with the provisioning window honored, and the stopped-server rebuild settling back to stopped — which also measures the previously-unverified assumption that Nova preserves power state across rebuild, on both drivers.
- Rebuild atomicity was measured on the libvirt driver (kolla 2025.1): the first GET after accept already shows the target ref with `task_state='rebuilding'`, and failure moves `REBUILD→ERROR` without presenting as settled.
- **Ironic behaviour is now measured too** (same cloud, sushy-tools virtual metal, 2026-08-06): accept flips ref+`task_state` atomically, `rebuilding`/`rebuild_spawning` stay visible through the ~2.5 min redeploy, and a failed redeploy settles `ERROR` **with the ref still on the target** — it never reverts, so a failed rebuild cannot re-trigger the destructive row. The node lands `deploy failed` and a plain rebuild to another image recovers both server and node with no operator action. One landmine found and documented: Nova microversion ≥ 2.93 breaks every Ironic rebuild (`reimage_boot_volume` is set unconditionally and the Ironic driver refuses it outright); our client pins 2.90 and now carries a comment forbidding a bump past 2.92. Residual for real hardware: deploy timing and firmware/BMC quirks, not the control-plane contract.


spjmurray (commented):
This is the right shape. One writer per fact, the decision taken from a fresh read, the marker gone from the schema, and liveness moved off the monitor onto the requeue timer. Net 450 lines lighter. And you went and measured the things the rest of us were guessing at, which is worth more than the diff is.

Six things before it goes in. The first two are regressions from main and are not about the rebuild argument at all.

### 1. Evacuation drops healthy servers out of `provisioned`

`R3` fires on **any** in-flight Nova rebuild, not just one we submitted. `nova evacuate` is implemented as a rebuild: `task_state=rebuilding`, and the image ref never moves, so it already equals `spec.Image.ID`.

So when a compute host dies and the operator evacuates, every server on it takes `R3`, gets `Active=Rebuilding` + `Healthy=Unknown`, and yields — re-written every 10s until the evacuation finishes. `provisioningStatus` flips `provisioned` → `provisioning` with nothing user-initiated behind it, and anything treating not-provisioned as not-ready acts on healthy servers.

On main that same read reached `reconcileServerImageConverged` with `rebuild == nil` and returned clean.

The R3 comment says "our own rebuild is mid-flight". That's the one thing a fresh read genuinely cannot tell you, and it's the honest cost of dropping the intent marker. Worth deciding deliberately rather than inheriting: either narrow the row, or accept that a foreign rebuild reads as ours.

### 2. Security group deletes fail for the whole rebuild window

`submitServerRebuild` and `R3` both always return `ErrYield`, so `Provision` never reaches `removeConsumedResourceReferences` (`provisioner.go:509`).

Change an image and drop a security group in the same PATCH, and region keeps its reference on the dropped group for the entire rebuild. A `DELETE` on that group is refused 403 "in use" until convergence — 30+ minutes on baremetal. Main's in-flight rows returned nil and released within one pass.

### 3. Two measurements of a failed baremetal deploy disagree

You measured (kolla + sushy-tools, 6 Aug): a failed redeploy settles Nova **ERROR** with the ref still on the target.

Jacob measured (dev + ironic-fake-control, 30 Jul, 3 of 3) in INST-1235: Nova stays **ACTIVE**, `task_state` empty, ref on the target, and only `powerState` reaches `Error`.

Both can't be right, and which one it is decides whether this PR fixes INST-1235 or ships it:

- If yours holds, `R3′` returns done, `provisioningStatus=provisioned` (correct — the reconciler did what was asked and can park), and the failure lands on `Active=Error` / `Healthy=Degraded`. Fine.
- If Jacob's holds, `R3′` returns done, `convertServerHealthStatus` maps ACTIVE to healthy, `setServerActive` maps it to Running, and `servers.CRASHED` at `provider.go:2587` is still an empty `REVIEW_ME`. The failure lands nowhere at all.

Cheap to settle: Jacob's suite is on #616, unmerged, and `server_faults_test.go` is on neither main nor this branch — so nothing in CI can catch it either way. Run that suite against this branch before merging.

Separately, and beyond this PR: a "deploy fail" injected by the fake-control sidecar and a real sushy-tools deploy failure are producing different Nova-visible outcomes. One of those fixtures isn't representative of real hardware, and it would be good to know which.

While you're in there, `servers.CRASHED` is one line from being resolved and INST-1235 names it directly.

### 4. The deleted tests take the safety net with them

Ten gone: five park tests in `reconcile_server_image_test.go`, and five in `client_v2_test.go` that pinned "an unrealized image never reads as provisioned at the API".

The one replacement, `TestReconcileServerImageConvergedErrorIsObserved`, asserts `require.NoError` on a converged ERROR server. So the suite now certifies the exact behaviour that's in dispute in (3), and no test anywhere asserts that a rebuild failure reaches a user-visible field. Whatever the answer to (3) turns out to be, it needs a test that would fail if someone undoes it.

### 5. `status.observed.error` is the wrong half of the old behaviour

Main produced a translated, actionable message on failure:

> "server entered an error state after a rebuild was issued; select another image or replace the server"

That's ours, provider-neutral, and tells the user what to do. This PR drops it for the generic "server is in an error state", and keeps Nova's raw fault code and text in `status.observed.error` instead.

That's the wrong way round on both counts:

- The raw Nova fault must never reach the API — a Nova fault code and message are OpenStack vocabulary, and putting them on the wire makes the provider part of our contract.
- The translated message is exactly what should reach it, and it's the bit that got deleted.

Which leaves `observed.error` with no legitimate reader. Nothing in `pkg/` reads it (the monitor README says as much), and `ServerV2Read` doesn't expose it — correctly. Operator detail belongs in the log stream, where provider vocabulary is fine. The user-facing reason belongs in a neutral message on the health condition. Both are derived at the moment of observation; neither needs Nova's fault struct persisted in a status field to get there.

If `observed` needs an error member so the wake has something to diff, make it the neutral form. The bounds-checked `*int32` code, the `at` timestamp and the "no detail visible vs code 0" distinction are all built for a reader that shouldn't exist. Status fields can't be withdrawn once shipped, so this is worth getting right now rather than later.

Same question for the rest of the region, less sharply: `observed.image` is an `ImageID`, which is ours, so that one's fine.

### 6. Smaller things

- **Commit prefixes.** All six are `feat(server):` / `refactor(server):` / `docs(server):`. We merge by rebase, so they land on main verbatim and we end up with two conventions in the history. Needs a reword.
- **#626 is subsumed** — same CRD, same `types.go`, same `client_v2.go`, same README, and `observed.image` replaces `observedImageID`. Worth closing. #605 looks like it overlaps `observedServerError` too.
- **The 2.90 microversion pin is a good catch**, but it's guarded by a comment. A dependabot bump or a tidy-up walks straight through it and every baremetal rebuild breaks. Worth a unit test asserting the string.
- **R4″ on upgrade.** Dropping `status.rebuild` from the schema means the apiserver prunes it, so a server parked on a failed rebuild whose ref sits off-target becomes indistinguishable from one never attempted. The first reconcile after the upgrade issues a rebuild nobody asked for. Worth checking whether any such server exists in dev/staging before this lands.
- **`R2` yields forever.** An unreadable image ref with a differing spec image yields on every pass with no terminal state — `provisioning` for the life of the server and no message. The deleted code documented that ordering as deliberate so an ERROR with an unreadable ref parked instead.

spjmurray:
Measurements answer it, thanks — particularly splitting the libvirt and Ironic paths, which is what dissolved point 3 rather than either of us arguing it. Two things left and I'm happy.

### 1. Fold the fixups

The twelve commits on top need folding into the series they fix — we merge by rebase, so they land on main verbatim. One correction to the plan: this PR isn't a draft, so there's no later moment to do it in.

### 2. Finish the sweep you started

`R3′` was one instance of "I can't verify this, so I'll report success". Two more are sitting in the same function:

- **`R2`** — an unreadable ref yields on every pass with no terminal state and no message. You've already acknowledged it. I'd take it here rather than as a follow-up: the park pattern you just built *is* the fix, so it's a few lines while the context is hot.

- **`R1`** — the `spec.Image == nil` guard is dead. `image` is in the CRD's `required` list, and it has been non-optional since the type was written in #71, so there's no legacy object that could carry a nil. The dead code isn't what bothers me — it's that it resolves to **done**. If it ever did fire, a server would report `provisioned` onto no image at all. Same mistake as `R3′`, so it wants the same answer: yield or park, not complete.

Neither is large, and doing all three together means nobody finds this shape a third time.

alex-emery:
Both taken.

**R2 and R1** park via the same `UserActionRequired` path as R3′, re-derived per pass. R2 keeps its log line and gains a user-visible message ("the provider cannot report the server's image, so the desired image cannot be verified; replace the server"). R1 keeps the guard but parks instead of completing ("the server specifies no image to converge onto; set an image in the specification") — dead in practice, honest if it ever fires. Tests pin both.

**Folded.** The series is now ten commits, each a feature chapter — the parks (R3′, the image-not-found submission park, R2, R1) all live in *Decide server rebuilds from a fresh provider read*; the neutral `errored` marker and the pure/bounded projection are born correctly in *Record provider facts*; the doc corrections sit with the commits whose docs they correct. Every commit builds, vets, and is gofmt-clean individually; the tip tree is byte-identical to the reviewed head plus the R1/R2 change.

spjmurray (approved):

PR #644, opened 2026-08-10
Publish the Volume API contract

pcriv:
## Summary

Clients and later handler work need a stable Region API contract for project-scoped block volumes.

This change publishes the v2 CRUD contract with network placement, class, requested capacity, and provider-observed capacity. Attachment-derived status stays out of the base resource because attachments have a separate API and lifecycle.

Handler behavior and quota admission remain in their dedicated changes so this contract stays independently reviewable.

## Validation

- make touch
- make license
- make validate
- make lint
- make generate
- make test-unit

Fixes [STO-129].


pkg/openapi/server.spec.yaml
```diff
@@ -3541,6 +3678,94 @@ components:
           $ref: 'https://raw.githubusercontent.com/unikorn-cloud/core/v1.17.1/pkg/openapi/common.spec.yaml#/components/schemas/resourceWriteMetadata'
         spec:
           $ref: '#/components/schemas/serverV2Spec'
+    volumeV2Phase:
```
pcriv: let's remove this for now

pkg/openapi/server.spec.yaml
```diff
@@ -3541,6 +3678,79 @@ components:
...
+      - capacityGiB
+      properties:
+        networkId:
+          description: The Network that determines the volume's Region and project placement.
+          allOf:
+          - $ref: '#/components/schemas/networkId'
+        volumeClassId:
+          description: The provider-neutral VolumeClass used to provision the volume.
+          type: string
+        capacityGiB:
```
nsricardor: this introduces spec.capacityGiB, while requested storage quantities elsewhere use sizeGiB, including file storage and image schemas; VolumeClass bounds also use minimumSizeGiB/maximumSizeGiB. Rename the requested field to sizeGiB

pkg/openapi/server.spec.yaml
```diff
@@ -1810,6 +1810,128 @@ paths:
           $ref: 'https://raw.githubusercontent.com/unikorn-cloud/core/v1.17.1/pkg/openapi/common.spec.yaml#/components/responses/forbiddenResponse'
         '500':
           $ref: 'https://raw.githubusercontent.com/unikorn-cloud/core/v1.17.1/pkg/openapi/common.spec.yaml#/components/responses/internalServerErrorResponse'
+  /api/v2/volumes:
+    description: Manages project-scoped block storage volumes anchored to a Network.
+    get:
+      description: List volumes visible to the caller, optionally narrowed by scope or placement.
```
nsricardor: The new Volume documentation uses “placement” as an informal synonym for Region/Network association in several places, including pkg/openapi/server.spec.yaml:1816, pkg/openapi/server.spec.yaml:3682, and pkg/openapi/server.spec.yaml:3728.

However, Placement is already a first-class Reservation API resource with /api/v2/placements and placementV2* components in the Reservation specification (https://github.com/nscaledev/reservation/blob/main/pkg/openapi/server.spec.yaml). Using the same term for ordinary Region/Network scoping could imply an integration with that resource.

pkg/handler/handler_v2.go
```diff
@@ -29,6 +29,26 @@ import (
 	"github.com/unikorn-cloud/region/pkg/openapi"
 )
 
+func (h *Handler) GetApiV2Volumes(w http.ResponseWriter, r *http.Request, params openapi.GetApiV2VolumesParams) {
+	openapi.Unimplemented{}.GetApiV2Volumes(w, r, params)
```
nsricardor, line 33: These endpoints currently return an undocumented 501 from openapi.Unimplemented. With the default runtime-schema-validation=true and runtime-schema-validation-panic=true, this causes a request-level panic and connection reset. Please hide them or declare the response.

pcriv: We can leave it as is, and merge this PR together with the Handlers implementation

nsricardor (approved):

PR #646, opened 2026-08-10
feat: add InfiniBand flavor metadata

Lumin343:
## Summary

Adds optional InfiniBand metadata to flavors: `infiniBand.portCount` on `FlavorMetadata` (CRD, `minimum: 1`), flowing through the OpenStack provider conversion and surfaced on the flavor read API as `ib.portCount`. Consumers use it to know how many IB ports a flavor exposes — the natural capability discriminator for InfiniBand-capable flavors.

## Scope change

This PR originally carried three changes (gate arming, a server references sub-resource, and flavor metadata) as the region-side surface for `ib-manager`. Following review (see discussion below), the gate-based approach is being dropped for InfiniBand in favour of **program-then-create** — decide placement, program the fabric off the pinned host + network, then create the Server — so the coordination machinery isn't needed:

- **Arming** (`Region.Spec.ReadinessGates`) — removed. It also contradicted the region-level-defaulting non-goal in the pre-create service design (nscaledev/uni-specifications#43).
- **Server references sub-resource** — removed. It only existed to anchor the gate mechanism to the Server, which the placeholder Server no longer needs to be.

Only the flavor metadata remains — useful regardless of the coordination mechanism, and the IB discriminator either way.

## Testing

- `go build ./...`, `go vet`, `gofmt` clean; conversion + provider unit tests pass.
- Zero codegen drift (deepcopy + OpenAPI + chart CRDs regenerate clean).


spjmurray:
Reviewed the reduced PR. Small and clean, and it does what the summary says. Findings, worst first.

### 1. `ib` vs `infiniBand` — the API surface is the odd one out

The CRD field is `infiniBand`. The Go types are `InfiniBand`, `InfiniBandSpec` and `types.InfiniBand`. The published API is `ib`, `ibSpec` and `IbSpec`.

GPU pairs up the whole way through (`gpu` / `GPUSpec` / `gpuSpec` / `GpuSpec`); this does not. The spec is published to `nscaledev/openapi` (#654), so of everything here it is the one thing that gets expensive to change after merge. Suggest `infiniBand` throughout, to match the CRD and the Go types.

### 2. The half that reads the operator's config is untested

The new test is at the conversion layer, covering `types.Flavor` → `openapi.Flavor`. The provider layer — `Region.Spec.Openstack.Compute.Flavors.Metadata[].infiniBand` → `types.Flavor` — has nothing.

`ConvertFlavors` is already exported through `export_test.go`, and `architecture_test.go:76` shows the pattern, so this is a handful of lines. It is also what Codecov is flagging at ~54% on `provider.go`, and in this instance the complaint is the right one: `applyFlavorMetadata` is the new code and only its architecture path is exercised today.

### 3. Commit message

`feat: add InfiniBand flavor metadata` is at odds with `docs/commit-messages.md` on two counts: the conventional-commit prefix, and the lower-case subject. `Add InfiniBand flavor metadata` fits.

### 4. Worth splitting the refactor from the feature

Extracting `applyFlavorMetadata` is a good change, and it is what lets the InfiniBand block land without deepening the `nolint:nestif`. But it is a behaviour-preserving refactor bundled into a feature commit. PRs here land by rebase rather than squash, so two commits read better on `main`: extract first, then add.

(I checked the extraction is in fact behaviour-preserving — `Baremetal` and `PinnedOnly` still only apply on a metadata hit, before and after.)

### 5. "ports (HCAs)" conflates two things

An HCA is the card; a card carries one or two ports. The doc comment on `PortCount` treats them as the same thing. If the consumer ends up programming per-port GUIDs then that distinction is the whole game — worth picking one and saying which.

### 6. No `minimum` on the OpenAPI `portCount`

The CRD enforces `minimum: 1`, the published schema does not. It is a read-only response so the stakes are low, but the schema is what consumers generate from.

---

One observation rather than a finding: as a capability discriminator only the *presence* of `infiniBand` matters — the count is payload nothing in this repo reads. That is fine, and it is exactly the arrangement GPU already has. It does mean `portCount` is operator-asserted with nothing checking it against the hardware, so it can drift silently. Same as the GPU vendor/model/memory numbers, so not a new problem — just worth knowing it is a hand-maintained number and not an observed one.

PR #647, opened 2026-08-10
Implement Volume API handlers

pcriv:
## Summary

- Add Volume create, list, get, update, and delete handlers.
- Derive Region, Identity, organization, and project scope from the validated Network.
- Store requested capacity, preserve immutable size and class, and reject deletion while attached.
- Add focused handler, lifecycle, and RBAC unit coverage.

## Deferred scope

- STO-131 adds attachment API projection and Server attachment behavior.
- STO-159 adds create-time quota admission. Until then, Volume creation does not enforce organization quota.
- STO-133 adds full KinD Volume lifecycle integration coverage.

## Validation

- `make touch`
- `make license`
- `make validate`
- `make lint`
- `make generate`
- `make test-unit`

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/pkg/handler/handler_v2.go
+++ b/pkg/handler/handler_v2.go
@@ -26,27 +26,75 @@ import (
 	"github.com/unikorn-cloud/region/pkg/handler/loadbalancer"
 	"github.com/unikorn-cloud/region/pkg/handler/sshcertificateauthority"
 	"github.com/unikorn-cloud/region/pkg/handler/storage"
+	"github.com/unikorn-cloud/region/pkg/handler/volume"
 	"github.com/unikorn-cloud/region/pkg/openapi"
 )
 
+func (h *Handler) volumeClient() *volume.Client {
+	return volume.New(h.ClientArgs)
+}
+
 func (h *Handler) GetApiV2Volumes(w http.ResponseWriter, r *http.Request, params openapi.GetApiV2VolumesParams) {
-	openapi.Unimplemented{}.GetApiV2Volumes(w, r, params)
+	result, err := h.volumeClient().ListV2(r.Context(), params)
+	if err != nil {
+		errors.HandleError(w, r, err)
+		return
+	}
+
+	util.WriteJSONResponse(w, r, http.StatusOK, result)
 }
 
 func (h *Handler) PostApiV2Volumes(w http.ResponseWriter, r *http.Request) {
-	openapi.Unimplemented{}.PostApiV2Volumes(w, r)
+	request := &openapi.VolumeV2Create{}
+
+	if err := util.ReadJSONBody(r, request); err != nil {
+		errors.HandleError(w, r, err)
+		return
+	}
+
+	result, err := h.volumeClient().CreateV2(r.Context(), request)
+	if err != nil {
+		errors.HandleError(w, r, err)
+		return
+	}
+
+	util.WriteJSONResponse(w, r, http.StatusAccepted, result)
 }
 
 func (h *Handler) GetApiV2VolumesVolumeID(w http.ResponseWriter, r *http.Request, volumeID openapi.VolumeIDParameter) {
-	openapi.Unimplemented{}.GetApiV2VolumesVolumeID(w, r, volumeID)
+	result, err := h.volumeClient().GetV2(r.Context(), volumeID)
+	if err != nil {
+		errors.HandleError(w, r, err)
+		return
+	}
+
+	util.WriteJSONResponse(w, r, http.StatusOK, result)
 }
 
 func (h *Handler) PutApiV2VolumesVolumeID(w http.ResponseWriter, r *http.Request, volumeID openapi.VolumeIDParameter) {
-	openapi.Unimplemented{}.PutApiV2VolumesVolumeID(w, r, volumeID)
+	request := &openapi.VolumeV2Update{}
+
+	if err := util.ReadJSONBody(r, request); err != nil {
+		errors.HandleError(w, r, err)
+		return
+	}
+
+	result, err := h.volumeClient().UpdateV2(r.Context(), volumeID, request)
+	if err != nil {
+		errors.HandleError(w, r, err)
+		return
+	}
+
+	util.WriteJSONResponse(w, r, http.StatusOK, result)
 }
 
 func (h *Handler) DeleteApiV2VolumesVolumeID(w http.ResponseWriter, r *http.Request, volumeID openapi.VolumeIDParameter) {
-	openapi.Unimplemented{}.DeleteApiV2VolumesVolumeID(w, r, volumeID)
+	if err := h.volumeClient().DeleteV2(r.Context(), volumeID); err != nil {
+		errors.HandleError(w, r, err)
+		return
+	}
+
+	w.WriteHeader(http.StatusAccepted)
 }
 
 func (h *Handler) GetApiV2Networks(w http.ResponseWriter, r *http.Request, params openapi.GetApiV2NetworksParams) {
--- a/pkg/handler/volume/README.md
+++ b/pkg/handler/volume/README.md
@@ -1,0 +1,20 @@
+# Volume
+
+`pkg/handler/volume` implements the public v2 Volume lifecycle API.
+
+Volume creation validates the selected Network and VolumeClass, derives Region,
+Identity, organization, and project scope from that Network, and persists a
+controller-finalized `Volume` CRD owned by the Network with deletion blocking.
+Requested size is stored as a binary quantity while the API remains whole-GiB.
+Quota admission is not part of this package. A later slice adds create-time
+quota admission.
+
+List and get expose project-scoped metadata, immutable Network/class/size, the
+standard lifecycle and health projection, and observed size when it is present
+in CRD status. Attachment-derived fields are not projected by this package.
+
+Updates replace mutable metadata and tags with optimistic locking while
+preserving Network, VolumeClass, size, claims, controller state, and any
+future allocation annotation. Delete is idempotent once started and rejects a
+Volume with an active attachment claim or resource reference until it is
+manually detached.
--- a/pkg/handler/volume/client.go
+++ b/pkg/handler/volume/client.go
@@ -1,0 +1,406 @@
+/*
+Copyright 2026 Nscale.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package volume
+
+import (
+	"cmp"
+	"context"
+	"fmt"
+	"math"
+	"slices"
+
+	coreconstants "github.com/unikorn-cloud/core/pkg/constants"
+	"github.com/unikorn-cloud/core/pkg/manager"
+	"github.com/unikorn-cloud/core/pkg/server/conversion"
+	"github.com/unikorn-cloud/core/pkg/server/errors"
+	coreutil "github.com/unikorn-cloud/core/pkg/server/util"
+	identitycommon "github.com/unikorn-cloud/identity/pkg/handler/common"
+	identityids "github.com/unikorn-cloud/identity/pkg/ids"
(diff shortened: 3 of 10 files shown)


pkg/handler/volume/client.go
```diff
@@ -0,0 +1,400 @@
...
+	}
+
+	resource, err := c.generate(ctx, organizationID, projectID, &openapi.VolumeV2Update{Metadata: request.Metadata}, network, request.Spec.VolumeClassId, size)
+	if err != nil {
+		return nil, err
+	}
+
+	resource.Finalizers = []string{coreconstants.Finalizer}
+
+	if err := c.Client.Create(ctx, resource); err != nil {
```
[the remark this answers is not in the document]
pcriv: Fixed in 3184f05. Volume CRDs now carry a Network owner reference with blockOwnerDeletion=true. Network deletion already runs as a foreground Identity deletion, so the Identity and provider credentials remain until the Network-dependent Volume completes finalizer-driven provider teardown. The STO-126 lifecycle also returns missing Identity/provider lookup errors and preserves the Volume finalizer instead of treating provider state as absent. Focused handler, provisioner, and deletion tests pass.

pkg/handler/volume/client.go
```diff
@@ -0,0 +1,405 @@
...
+type Client struct {
+	common.ClientArgs
+}
+
+// New creates a Volume client.
+func New(args common.ClientArgs) *Client {
+	return &Client{ClientArgs: args}
+}
+
+func sizeGiB(quantity resource.Quantity) int64 {
```
squaremo, line 66: This is only size in GiB if the quantity is denominated in bytes in the first place; so it'd be easy to misuse if not paying too much attention. But, since it's local to the package, I think we can excuse this one (though maybe make the assumption obvious with the name, or a comment).

pkg/handler/handler_v2.go
```diff
@@ -26,27 +26,71 @@ import (
 	"github.com/unikorn-cloud/region/pkg/handler/loadbalancer"
 	"github.com/unikorn-cloud/region/pkg/handler/sshcertificateauthority"
 	"github.com/unikorn-cloud/region/pkg/handler/storage"
+	"github.com/unikorn-cloud/region/pkg/handler/volume"
 	"github.com/unikorn-cloud/region/pkg/openapi"
 )
 
 func (h *Handler) GetApiV2Volumes(w http.ResponseWriter, r *http.Request, params openapi.GetApiV2VolumesParams) {
-	openapi.Unimplemented{}.GetApiV2Volumes(w, r, params)
+	result, err := volume.New(h.ClientArgs).ListV2(r.Context(), params)
```
squaremo: The existing handler methods use a helper to create the client. They could reasonably cache the client, I think, though it's not done so far as I can see (so the helper is just there to hide the cache-or-not-cache decision).
You can agree or disagree with the value of the helper, but it is worth keeping to the convention at least.

pkg/handler/volume/client.go
```diff
@@ -0,0 +1,405 @@
...
+	result, err := c.GetV2Raw(ctx, volumeID.String())
+	if err != nil {
+		return nil, err
+	}
+
+	return convertV2(result)
+}
+
+func requestedSize(sizeGiB int64) (resource.Quantity, error) {
+	if sizeGiB > math.MaxInt64/(1<<30) {
```
squaremo, line 201: This feels a bit like overkill, but we might not detect an overflow otherwise, so fair enough (https://go.dev/play/p/63Us1_xOM0W?v=goprev)

spjmurray (commented):
Reviewed against `pkg/handler/securitygroup/client_v2.go` (closest sibling: v2-only, Network-owned child), `pkg/handler/storage/client.go`, and the platform spec v3.23.

Structurally this is right, and nothing below is architectural. It is a faithful v2 handler: scope derived from the parent Network, selector-prefiltered list then tag filter then per-item RBAC, owner reference with `BlockOwnerDeletion`, optimistic-lock patch, RBAC in the client rather than the transport layer, and the clusterrole rule picked correctly.

Two things that look inconsistent with the rest of the repo are actually the other way round, so please don't let anyone talk you out of them:

- **202 on create is correct.** §7.1 keys the code off whether a controller has to drive the resource to a terminal state, which Cinder provisioning does. The 201s on networks, security groups, load balancers and filestorage predate the spec and are frozen for client compatibility.
- **Setting the finalizer in the handler is correct.** §7.2 Create: "Add the controller's finalizer to the resource metadata before any controller is able to observe the resource", with the reasoning in §7.6. Relying on core's reconciler to add it later leaves exactly the create-then-delete-before-first-reconcile window the spec closes.

200 on update also holds up: the mutable set is name, description and tags, none of which reaches Cinder, so it is fully effected in the request cycle.

Two comments that don't anchor to a line:

**Tests.** Every other v2 resource has a `test/api/suites/<x>_test.go` and an `endpoints.go` entry; volumes has neither. The unit coverage here is good (fake client, mocked providers, table-driven), but there are also no RBAC-denial cases — securitygroup and storage both keep a `rbac_test.go` covering unauthorized org/project and cross-project reads.

**Commits.** "Link Volumes to their Network owner" is a 5-line fixup to a handler this same PR introduces. This repo rebases rather than squashes, so it lands on main verbatim — worth folding into "Implement Volume API handlers".

pkg/handler/volume/client.go
```diff
@@ -0,0 +1,405 @@
...
+func New(args common.ClientArgs) *Client {
+	return &Client{ClientArgs: args}
+}
+
+func sizeGiB(quantity resource.Quantity) int64 {
+	return quantity.Value() / (1 << 30)
+}
+
+func convertV2(in *regionv1.Volume) (*openapi.VolumeV2Read, error) {
+	regionID, err := regionids.ParseRegionID(in.Labels[constants.RegionLabel])
```
spjmurray: Parsing the region label inline duplicates `regionIDFromLabels` in `pkg/apis/unikorn/v1alpha1/helpers.go`, down to a bespoke error string. `Server.RegionID()` is the established shape and `Volume.NetworkID()` already sits next to it, so a `Volume.RegionID()` accessor would keep this recovery fail-closed in one place. `pkg/handler/README.md` names this as the convention.

pkg/handler/volume/client.go
```diff
@@ -0,0 +1,405 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	resource, err := c.generate(ctx, organizationID, projectID, &openapi.VolumeV2Update{Metadata: request.Metadata}, network, request.Spec.VolumeClassId, size)
+	if err != nil {
+		return nil, err
+	}
+
+	resource.Finalizers = []string{coreconstants.Finalizer}
```
spjmurray, line 302: Right call, but it is the only place in `pkg/handler` that does it, so it reads as a mistake to anyone going off local precedent. A one-line comment citing §7.2 ("add the controller's finalizer before any controller is able to observe the resource") would stop a future reviewer deleting it.

pkg/handler/volume/client.go
```diff
@@ -0,0 +1,405 @@
...
+
+	if err := c.Client.Create(ctx, resource); err != nil {
```
spjmurray, line 304: §7.6 wants a soft quota reservation here before the 202, and a 403 when the organisation is short. Deferring it is a reasonable slice boundary, but the effect until it lands is that volumes can be created past an organisation's limit, so it is worth stating on the PR rather than only in the package README.

pkg/handler/volume/client.go
```diff
@@ -0,0 +1,405 @@
...
+	if err := rbac.AllowProjectScopeReader(ctx, endpoint, identityapi.Delete, resource); err != nil {
+		return err
+	}
+
+	if resource.DeletionTimestamp != nil {
+		return nil
+	}
+
+	if resource.Spec.ClaimRef != nil || len(manager.GetResourceReferences(resource)) != 0 {
+		return errors.HTTPConflict().WithValues("reason", "volume is attached and must be detached before deletion")
```
spjmurray: `errors.HTTPConflict()` takes no description — core hard-codes the body as "the requested resource already exists". So a caller trying to delete an attached volume gets a 409 telling them it already exists, and the actual reason only reaches the server log through `WithValues`. The spec's error rule is that descriptions must carry what the user needs to take corrective action, and here that is "detach it first".

The existing idiom is `errors.HTTPForbidden("security group is in use and cannot be deleted")`, and the delete operation already declares 403, so switching needs no contract change. Otherwise a describable conflict helper in core.

pkg/handler/handler_v2.go
spjmurray: The other five v2 resources each have a small `h.networkClient()` / `h.securityGroupClient()` / `h.loadBalancerClient()` constructor; this repeats `volume.New(h.ClientArgs)` across all five handlers. A `h.volumeClient()` would match.

pkg/handler/volume/README.md
```diff
@@ -0,0 +1,20 @@
+# Volume
+
+`pkg/handler/volume` implements the public v2 Volume lifecycle API.
+
+Volume creation validates the selected Network and VolumeClass, derives Region,
+Identity, organization, and project scope from that Network, and persists a
+controller-finalized `Volume` CRD owned by the Network with deletion blocking.
+Requested size is stored as a binary quantity while the API remains whole-GiB.
+Quota allocation is deliberately not performed here yet; STO-159 adds that
```
spjmurray: This repo is public, so the Linear ID shouldn't go in committed docs. Describing the unblocking condition — "a later slice adds create-time quota admission" — carries the same information without the internal reference.

pkg/handler/volume/client.go
pcriv: Added `Volume.RegionID()` and routed the read conversion through the shared fail-closed label helper.

pkg/handler/volume/client.go
pcriv: Added the §7.2 rationale above finalizer installation.

pkg/handler/volume/client.go
pcriv: Updated the PR body to state that create-time quota admission is deferred to STO-159 and is not enforced in this slice.

pkg/handler/volume/client.go
pcriv: Changed attached deletion to the established 403 Forbidden pattern with an actionable detach-first message, and updated the tests.

pkg/handler/handler_v2.go
pcriv: Added `volumeClient()` and reused it across all five Volume handlers.

pkg/handler/volume/README.md
pcriv: Removed the internal ticket reference from the committed package documentation.

pcriv:
Addressed the review: added focused Volume RBAC-denial coverage, folded the Network owner-reference change into the implementation commit, and documented the deferred quota behavior in the PR body. Full KinD lifecycle coverage remains in STO-133.

pcriv:
@codex review

squaremo:
> * Relying on core's reconciler to add it later leaves exactly the create-then-delete-before-first-reconcile window the spec closes.

Typically, the pattern is that the controller adds the finaliser (and persists the change) before it does anything that might need to be fixed up in finalisation. What is it that is done between the resource being created by the handler, and the controller seeing it, that would need to be fixed up? (Is it claiming allocations?)

squaremo (dismissed):
Looks fine to me, too. Nice to see this overall feature broken down into smallish, self-contained changes.

pkg/handler/volume/client.go
pcriv: Agreed. The helper is package-local and both callers pass Volume storage quantities, so I am keeping it local as-is.

pcriv:
@squaremo Nothing is created outside Kubernetes before the first reconcile in this slice; quota admission is deferred to STO-159. The handler-owned finalizer preserves the §7.2 invariant now and becomes necessary when create-time quota reservation is added: an immediate delete must leave the CRD observable so deprovisioning can release that reservation. Provider deletion is idempotent, so a delete before the first provision pass is safe.

pkg/handler/handler_v2.go
pcriv: Implemented in `567244a`: all five Volume handlers now use `h.volumeClient()`.

pkg/handler/volume/client.go
pcriv: Kept the guard because it prevents `int64` overflow before converting GiB to bytes.

squaremo (dismissed):

squaremo (approved):

PR #649, opened 2026-08-11
Add VolumeClass flavor allowlist

pcriv:
VolumeClass inventory cannot describe which Region flavors work with a
storage class. Consumers cannot distinguish a restricted class from one
that works with every flavor.

This change adds an extensible operator-authored `supportedFlavors`
selector to Region VolumeClass metadata. Its initial `ids` condition is a
typed Flavor allowlist. OpenStack conversion resolves those IDs into the
flat provider-neutral and public `supportedFlavorIds` inventory field.

An omitted selector and omitted or empty IDs remain unrestricted. CRD
admission validates only canonical UUID shape and uniqueness because live
provider lookups do not belong there.

Attachment validation and enforcement remain with STO-168.

Tested:
- make touch
- make license
- make validate
- make lint
- make generate
- make test-unit

Fixes: STO-167


pkg/apis/unikorn/v1alpha1/types.go
```diff
@@ -342,6 +342,10 @@ type VolumeClassMetadata struct {
 	// ID is the immutable provider identifier for the volume class. For OpenStack,
 	// this is the Cinder volume type ID.
 	ID string `json:"id"`
+	// SupportedFlavorIDs optionally restricts this volume class to the listed
+	// Region flavors. Nil or empty means no compatibility restriction.
+	// +listType=set
+	SupportedFlavorIDs []regionids.FlavorID `json:"supportedFlavorIds,omitempty"`
```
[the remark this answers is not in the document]
pcriv: Fixed in 15fcba4. supportedFlavorIds items now require canonical lowercase, hyphenated UUID spelling through native OpenAPI type, format, and pattern validation, while the list retains set semantics. Regression cases cover uppercase and unhyphenated spellings of an existing UUID. ProviderContractVerification passes against Kubernetes 1.35.

pkg/apis/unikorn/v1alpha1/types.go
```diff
@@ -342,6 +342,13 @@ type VolumeClassMetadata struct {
 	// ID is the immutable provider identifier for the volume class. For OpenStack,
 	// this is the Cinder volume type ID.
 	ID string `json:"id"`
+	// SupportedFlavorIDs optionally restricts this volume class to the listed
+	// Region flavors. Nil or empty means no compatibility restriction.
+	// +kubebuilder:validation:items:Type=string
+	// +kubebuilder:validation:items:Format=uuid
```
nsricardor: The `items:Type=string` and `items:Format=uuid` markers here are no-ops. `regionids.FlavorID`
already carries `+kubebuilder:validation:Type=string` and `Format=uuid` on its type declaration
(pkg/ids/ids.go:28-29), and controller-gen propagates those to slice items.

I verified by deleting both lines and re-running controller-gen — the generated CRD was
identical except for the `pattern` line, so only the `items:Pattern` marker is actually doing
anything. Suggest dropping the two redundant markers to keep the "validation markers live on
the ID type" convention intact.

pkg/apis/unikorn/v1alpha1/types.go
```diff
@@ -342,6 +342,13 @@ type VolumeClassMetadata struct {
...
+	// +kubebuilder:validation:items:Pattern=`^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$`
+	// +listType=set
+	SupportedFlavorIDs []regionids.FlavorID `json:"supportedFlavorIds,omitempty"`
```
nsricardor: A flat ID list closes the door on non-ID selection rules. Every other selection point in this
CRD goes through a `*Selector` struct (`FlavorSelector`, `VolumeClassSelector`,
`NetworkSelector`) precisely so new conditions can be added without a breaking schema change.
Since this field hasn't shipped yet, it's free to change now and breaking later — e.g. if we
ever want to select all VM flavors rather than enumerate them.

  type VolumeClassMetadata struct {
      ID string `json:"id"`
      // SupportedFlavors optionally restricts this volume class to a subset of Region
      // flavors. Unlike VolumeClassSelector, which is fail-closed, an undefined selector
      // means no compatibility restriction.
      SupportedFlavors *VolumeClassFlavorSelector `json:"supportedFlavors,omitempty"`
      ...
  }

  // VolumeClassFlavorSelector selects the Region flavors a volume class is compatible with.
  type VolumeClassFlavorSelector struct {
      // IDs is an explicit list of allowed flavor IDs. If not specified, all flavors are
      // considered.
      // +listType=set
      IDs []string `json:"ids,omitempty"`
  }

Keep the public API field flat. `supportedFlavorIds: []flavorId` in the OpenAPI is the resolved
output — clients shouldn't evaluate selector rules. So this is a CRD-only change; the API
contract and `types.VolumeClass` stay as they are, and don't churn when the rules get richer.

pkg/apis/unikorn/v1alpha1/types.go
pcriv: Addressed in a581adb. Removed the field-level items:Type and items:Format markers. make generate confirms the generated CRD is unchanged because regionids.FlavorID already supplies both constraints; the field retains only its canonical-spelling pattern and set semantics. The full repository checklist passes.

pcriv: Correction after CI verification: items:Format was redundant and remains removed, but items:Type is required. A fresh controller-gen invocation failed without it with `must apply pattern to a textual value, found type ""` because item markers are applied before the imported FlavorID schema is resolved. Commit 73fef11 retains items:Type, removes only items:Format, and passes the exact CRD generator plus the full local checklist.

pkg/apis/unikorn/v1alpha1/types.go
pcriv: Implemented in 48e30a5. Region VolumeClass metadata now uses a dedicated `supportedFlavors` selector with typed, unique canonical UUIDs under `ids`; OpenStack resolves those IDs into the unchanged flat provider-neutral and public `supportedFlavorIds` field. Omitted selectors and omitted/empty IDs remain unrestricted. Generated CRDs, Helm schema/examples, tests, and package docs are updated. STO-167 and the PR description now record this split input/output contract.

nsricardor (approved):

PR #658, opened 2026-08-18
Enforce Volume capacity quota at API creation

pcriv:
## What changed

- Allocate requested Volume capacity in bytes against the Identity `volumes` quota before persisting a Volume.
- Compensate by deleting the allocation if the Volume write fails.
- Cover the allocation request and persisted allocation annotation; update lifecycle documentation.

## Why

Volume creation previously persisted the resource without quota admission even though the controller released allocation metadata during deletion.

## Validation

- `go test ./pkg/handler/volume ./pkg/provisioners/managers/volume`

### DIFF
--- a/pkg/handler/volume/client.go
+++ b/pkg/handler/volume/client.go
@@ -27,7 +27,9 @@ import (
 	"github.com/unikorn-cloud/core/pkg/manager"
 	"github.com/unikorn-cloud/core/pkg/server/conversion"
 	"github.com/unikorn-cloud/core/pkg/server/errors"
+	"github.com/unikorn-cloud/core/pkg/server/saga"
 	coreutil "github.com/unikorn-cloud/core/pkg/server/util"
+	identityclient "github.com/unikorn-cloud/identity/pkg/client"
 	identitycommon "github.com/unikorn-cloud/identity/pkg/handler/common"
 	identityids "github.com/unikorn-cloud/identity/pkg/ids"
 	identityapi "github.com/unikorn-cloud/identity/pkg/openapi"
@@ -263,6 +265,37 @@ func (c *Client) generate(ctx context.Context, organizationID identityids.Organi
 	return out, nil
 }
 
+type createSaga struct {
+	client *Client
+	volume *regionv1.Volume
+}
+
+func (s *createSaga) createAllocation(ctx context.Context) error {
+	return identityclient.NewAllocations(s.client.Client, s.client.Identity).Create(ctx, s.volume, identityapi.ResourceAllocationList{{
+		Kind:      "volumes",
+		Committed: int(s.volume.Spec.Size.Value()),
+	}})
+}
+
+func (s *createSaga) deleteAllocation(ctx context.Context) error {
+	return identityclient.NewAllocations(s.client.Client, s.client.Identity).Delete(ctx, s.volume)
+}
+
+func (s *createSaga) createVolume(ctx context.Context) error {
+	if err := s.client.Client.Create(ctx, s.volume); err != nil {
+		return fmt.Errorf("%w: unable to create volume", err)
+	}
+
+	return nil
+}
+
+func (s *createSaga) Actions() []saga.Action {
+	return []saga.Action{
+		saga.NewAction("create quota allocation", s.createAllocation, s.deleteAllocation),
+		saga.NewAction("create volume", s.createVolume, nil),
+	}
+}
+
 // CreateV2 creates a v2 Volume from Network-derived scope.
 func (c *Client) CreateV2(ctx context.Context, request *openapi.VolumeV2Create) (*openapi.VolumeV2Read, error) {
 	network, err := network.New(c.ClientArgs).GetV2Raw(ctx, request.Spec.NetworkId.String())
@@ -301,8 +334,8 @@ func (c *Client) CreateV2(ctx context.Context, request *openapi.VolumeV2Create) 
 	// Add the finalizer before the controller can observe the resource, as required by specification §7.2.
 	resource.Finalizers = []string{coreconstants.Finalizer}
 
-	if err := c.Client.Create(ctx, resource); err != nil {
-		return nil, fmt.Errorf("%w: unable to create volume", err)
+	if err := saga.Run(ctx, &createSaga{client: c, volume: resource}); err != nil {
+		return nil, err
 	}
 
 	return convertV2(resource)
--- a/pkg/handler/volume/client_test.go
+++ b/pkg/handler/volume/client_test.go
@@ -19,6 +19,7 @@ package volume_test
 import (
 	"context"
 	"math"
+	"net/http"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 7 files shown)


pcriv:
@codex review

pkg/handler/volume/client.go
```diff
@@ -263,6 +265,37 @@ func (c *Client) generate(ctx context.Context, organizationID identityids.Organi
 	return out, nil
 }
 
+type createSaga struct {
```
pcriv, line 268: **Standards (judgement call):** `createSaga` here only wraps the final two steps (allocate + persist), unlike the richer saga shape in `network/client_v2.go` and `storage/saga.go`, which fold validation/generation into the saga via a factory. Likely fine given Volume's simpler flow, but it's a real structural divergence from the sibling pattern worth a second look.

pkg/handler/volume/client.go
```diff
@@ -263,6 +265,37 @@ func (c *Client) generate(ctx context.Context, organizationID identityids.Organi
...
+	client *Client
+	volume *regionv1.Volume
+}
+
+func (s *createSaga) createAllocation(ctx context.Context) error {
+	return identityclient.NewAllocations(s.client.Client, s.client.Identity).Create(ctx, s.volume, identityapi.ResourceAllocationList{{
```
pcriv, line 274: **Standards (judgement calls):**
- The `identityapi.ResourceAllocationList{{Kind, Committed}}` + `identityclient.NewAllocations(...).Create/Delete` shape is now near-verbatim across three handler packages (network, storage, volume) — Duplicated Code, though this extends existing precedent rather than introducing it.
- `createAllocation`/`deleteAllocation` return raw unwrapped errors, unlike `storage/saga.go`'s `wrapAllocationError`. Matches `network`'s convention (which also doesn't wrap), so the siblings already disagree with each other here — not a clear violation, just inconsistent across the codebase.

pkg/handler/volume/client.go
```diff
@@ -301,8 +334,8 @@ func (c *Client) CreateV2(ctx context.Context, request *openapi.VolumeV2Create)
 	// Add the finalizer before the controller can observe the resource, as required by specification §7.2.
 	resource.Finalizers = []string{coreconstants.Finalizer}
 
-	if err := c.Client.Create(ctx, resource); err != nil {
-		return nil, fmt.Errorf("%w: unable to create volume", err)
+	if err := saga.Run(ctx, &createSaga{client: c, volume: resource}); err != nil {
```
pcriv, line 337: **Spec (lower-confidence, systemic):** the platform spec's §7.6 describes a two-phase model — handler makes a soft reservation, controller later promotes it to a committed allocation. Here `Committed` is set directly at handler-create time, and there's no `Promote` step/method anywhere in the identity client. That said, this matches three already-merged sibling handlers (network, loadbalancer, storage) exactly, so it's a repo-wide pattern question rather than a defect unique to this PR — flagging since this is the kind of change that could have closed the gap.

pkg/handler/volume/client_test.go
```diff
@@ -146,14 +160,35 @@ func expectVolumeClasses(ctrl *gomock.Controller, providers *mockproviders.MockP
...
+				JSON201: &identityapi.AllocationRead{Metadata: coreapi.ProjectScopedResourceReadMetadata{
+					Id:             "88888888-8888-4888-a888-888888888888",
+					OrganizationId: testOrganizationID,
+					ProjectId:      testProjectID,
+				}},
+			}, nil
+		})
+}
+
 func TestCreateV2DerivesScopeAndPersistsCapacity(t *testing.T) {
```
pcriv, line 203: **Spec:** the PR description says this covers "compensate by deleting the allocation if the Volume write fails," but no test here forces `Client.Create` to fail *after* `createAllocation` succeeds to verify the compensating `deleteAllocation` actually runs. Worth adding a case (e.g. object already exists / conflicting resource) that asserts the allocation-delete call happens on a downstream persist failure.

pkg/handler/volume/client_test.go
```diff
@@ -190,7 +225,7 @@ func TestCreateV2DerivesScopeAndPersistsCapacity(t *testing.T) {
 		Name:               testNetworkID,
 		BlockOwnerDeletion: ptr.To(true),
 	}}, stored.OwnerReferences)
-	require.NotContains(t, stored.Annotations, coreconstants.AllocationAnnotation)
+	require.Equal(t, "88888888-8888-4888-a888-888888888888", stored.Annotations[coreconstants.AllocationAnnotation])
 }
 
 func TestCreateV2ValidatesVolumeClass(t *testing.T) {
```
pcriv, line 317: **Spec:** no test exercises Identity rejecting the allocation (non-201 response) to confirm the Volume is never persisted in that case — only happy-path 201s are stubbed via `expectAllocationCreate`. Given this PR's whole point is enforcing quota admission, a quota-exceeded rejection path seems like the most important case to cover.

pcriv (commented):
Two-axis review (Standards + Spec). Left inline notes below; nothing here blocks merge.

pkg/handler/volume/client.go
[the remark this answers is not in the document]
pcriv: Fixed in 0cec911. `TestVolumeV2Handlers` now provides the Identity allocation mock required by the new create flow. `make test-unit` passes.

pkg/handler/volume/client.go
pcriv: Investigated and kept the compact saga. Validation and object generation occur before either side effect; the saga only needs to compensate the allocation when persistence fails. This matches the package's best-effort multi-operation semantics without adding no-op compensation.

pkg/handler/volume/client.go
pcriv: Investigated and intentionally left unchanged. The allocation shape and raw error propagation match existing Network behavior; the sibling handlers differ enough that a shared helper or wrapper policy would be a cross-handler refactor, not a Volume-specific fix.

pkg/handler/volume/client.go
pcriv: Confirmed this is a repository-wide Identity contract gap rather than a Volume-specific divergence: Network, LoadBalancer, and Storage also create committed allocations, and the current Identity client exposes no promote operation. Left unchanged for a dedicated cross-service decision.

pkg/handler/volume/client_test.go
pcriv: Fixed in 692fa9b. `TestCreateV2PersistenceFailureDeletesAllocation` forces the Volume create to fail after a successful allocation and asserts the compensating allocation DELETE.

pkg/handler/volume/client_test.go
pcriv: Fixed in 692fa9b. `TestCreateV2QuotaRejectionDoesNotPersistVolume` makes Identity return 409 and asserts the handler returns the conflict and creates no Volume.

pkg/handler/volume/client.go
squaremo: Maybe log this for follow-up? It'd be a modest, refactor-only change, so easy to review.

squaremo (approved):
- The shape of this looks right to me -- using a saga for the bits that may need to have compensating transactions is the right, conventional pattern.
 - The tests cover the useful scenarios (undoing after problems).
 - I am relying on the machine review to pick up errors in logic.

pkg/handler/volume/client.go
pcriv: Good one, I'll create a ticket for it!

PR #662, opened 2026-08-24
Support terminal provider-create gate blocks

Lumin343:
## What

Extends the Server provider-create gate so a satisfier can report a gate as **blocked**, and optionally **terminal**. A terminal block tells the provisioner the gate will never be satisfied, so provider-create **fails** instead of holding (yielding) indefinitely.

This unblocks the fabric-manager flow (INST-640): when p_key allocation cannot succeed (e.g. pool exhausted), the fabric-manager reports the Server gate as terminally blocked and the server create fails cleanly rather than hanging.

## Changes

- `serverProviderCreateGateAction` gains optional `status` (`True`/`False`, default `True`) and `terminal` (boolean; only valid with `status: False`).
- `ServerProviderCreateGateStatus` records `terminal`; `Server` exposes `TerminalProviderCreateGate()`.
- `SatisfyProviderCreateGate` parses/validates the new fields (rejects `terminal` without `status: False`).
- The server provisioner fails on a terminal gate rather than yielding.
- Regenerated openapi (`types.go`, `schema.go`) + CRD; added helper/handler/provisioner tests.

## Testing

- `make lint` clean for the diff (pre-existing `nolintlint` on untouched `GetEventRecorderFor` is a local go1.26 vs CI go1.25 `staticcheck` artifact).
- `go build ./...`, `go vet`, and gate tests pass.

Linear: INST-1369


Lumin343:
Addressed the review:

- **Documentation gap** (the actionable item): updated `pkg/apis/unikorn/v1alpha1/README.md` to document the new `status`/`terminal` fields and the fail-vs-hold lifecycle — unsatisfied or `False,terminal=false` holds (yield), `False,terminal=true` fails provider-create with the redacted reason, all `True` proceeds. This satisfies the documentation-maintenance contract in CLAUDE.md for a lifecycle-semantics change.
- **PR description inventory**: correct — `provisioner_retry_test.go` had a one-line fixture update (passing `terminal: false` to the extended `ProviderCreateGateStatusWrite`), part of updating all call sites of that helper.
- **Codecov patch %**: the uncovered lines are error-return branches in `client_v2.go` (RBAC/actor failures); the terminal validation, persistence, and provisioner fail-vs-yield paths are covered by the helper/handler/provisioner tests. Not a required check.

The earlier `mergeable=BLOCKED` was just the `Integration` check re-running after I updated the branch onto `main`.

spjmurray:
Coming back to this having read the wider trail, including the reservation-side program-then-create work that was built out and then cancelled. That changes my view of the shape argument I made on #646: the alternative as actually implemented was a new CRD, a readiness reconciler, capability finalizers and a validating admission policy — a larger coordination mechanism than the gate, just in a different repo. So I'm not re-opening that. Picking the smaller of the two was reasonable and I was arguing from half the picture.

The findings below hold either way.

### 1. A caller can now wedge or kill a running server

`providerCreateGateActionState` accepts `Closed` and `Locked` unconditionally — there is no check that the gate was previously `Open`, and no check that the provider server already exists. `Provision` calls `blockUntilDependenciesReady` on every reconcile and the gate check lives inside it, with no already-created guard. `Server` is a constant-reconcile resource, so that runs on every pass for the life of the object.

So any holder of `region:servers:v2/provider-create-gates:update` can report `Closed` on a long-running server and hang it indefinitely, or `Locked` and fail it outright. On `main` today that is impossible: the endpoint can only ever write `True`, which is monotonic and therefore safe to re-evaluate forever. This PR removes that property without replacing it.

Two ways out, not exclusive:

- guard the gate check on the provider server not yet existing, so gates stop being live input once the create has happened;
- reject `Closed`/`Locked` on a gate that has already been `Open` in the current attempt.

### 2. This contradicts the design doc

nscaledev/uni-specifications#43 is explicit: *"service callers cannot set a gate back to `False` or `Unknown`"*, and *"Region owns resets if the server controller needs pre-create work to run again"*. Reporting `Closed` is precisely a caller-driven reset.

That doc has been open since June while the mechanism was implemented, argued over, reversed and now extended, and it also still describes the references subresource that was dropped from #646. It would be worth updating and landing it ahead of this change, so the next reviewer starts from something true — the disagreement on #646 happened largely because the shared design context was not in the shared place.

### 3. The upgrade note describes a gentler failure than the one that happens

`state` is `required` with an enum, the Go field has no `omitempty`, and there is no CRD default. A gate persisted by the current code has `status: "True"` and no `state` key, so it decodes as `""` and re-serialises as `"state": ""`, which fails both `required` and the enum.

That is not "waits for its satisfier to re-report". It is every status write on that Server being rejected by the apiserver until someone hand-patches it — and the provisioner and monitor both write status, so they will get there long before any satisfier does.

Adding `+kubebuilder:default=Closed` alongside the enum marker makes the README's claim actually true: structural-schema defaulting is applied when the stored object is decoded, so old entries read back as `Closed` and subsequent writes validate.

The blast radius today looks like nothing — there are no gated servers in production — so this may fairly be a documentation fix rather than a code one. But as written the note reassures about a case that would in fact break.

### 4. The read model did not follow the write model

`status.remainingProviderCreateGates` is still a list of names, so `Closed` and `Locked` are indistinguishable to every consumer. A restarted satisfier cannot tell that it already reported `Locked` and will redo the work; a caller cannot tell "still being worked" from "will never happen" except by string-matching the error on the Available condition. If the write side gains a state, the read side should expose it.

### 5. `Locked` has no reset path

`ProviderCreateGatesReset` has exactly one caller, in `deleteFailedProviderServer`, which is only reached when a provider server exists and failed. A `Locked` gate fails *before* provider create, so there is no provider server, no retry, and no reset. The server sits errored with the gate `Locked` indefinitely unless the satisfier volunteers a fresh report — which, per (4), it has no way to know is needed.

If that is the intent, it is worth saying so. As it stands the README's "a retry resets every gate back to `Closed`" does not apply to the one state that most needs it.

### 6. Reason vocabulary and untrusted text

`ErrProviderCreateGateLocked` puts another service's free-text `reason`/`message` into the error that surfaces on the Server. Two things worth checking: that it lands as the generic `Errored` Available reason with the detail in the message rather than as a bespoke reason — the provisioning reason vocabulary is a closed, core-owned set — and that the message is length-bounded so a chatty satisfier cannot bloat the stored object.

PR #663, opened 2026-08-24
Expose NFS policy API fields

vinimdocarmo:
## Summary

- expose optional non-nullable `posixAcl` and `atimeUpdateIntervalSeconds` fields across File Storage NFS create, update, and read models
- materialize `false` and `0` on create while preserving omitted stored values on update
- reject null and out-of-range atime values through the OpenAPI contract without changing `rootSquash` or adding provider-specific behavior
- add OpenAPI, handler, CRD, and typed integration coverage plus the approved API/package documentation

## Why

Customers need a provider-neutral API contract for POSIX ACL and atime desired state before customer clients can manage these settings. This PR builds on the presence-aware persisted fields from #661.

Linear: [STO-183](https://linear.app/nscale-workspace/issue/STO-183/uni-region-expose-the-public-file-storage-api-contract)

## Impact

This publishes the Region code contract but makes no deployment rollout changes. Customer-facing deployment remains gated by the approved adoption and release sequence.

## Validation

- `make touch`
- `make license`
- `make validate`
- `make lint`
- `make generate`
- `make test-unit`
- `go test -tags=integration ./test/api/... -run '^$' -count=1`
- two-axis standards/spec review with no findings

`make validate-docs` is currently unavailable because the pinned `uni-core` module does not provide `github.com/unikorn-cloud/core/hack/docs`.

## Stack

Depends on #661.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback</a></sub>


pcriv (approved):

pkg/openapi/server.spec.yaml
```diff
@@ -3793,6 +3793,25 @@ components:
...
+            Whether extended POSIX ACL support is enabled. On create, omission resolves to false.
+            On update, omission preserves the current setting. Explicit null is invalid.
+
+            Enabling POSIX ACLs may reduce metadata performance. Extended POSIX ACLs must be managed over
+            NFSv3. Disabling this option does not remove existing ACLs; they may remain enforced.
+          type: boolean
+        atimeUpdateIntervalSeconds:
+          description: |
+            On create, omission resolves to 0. On update, omission preserves the current setting.
+            Explicit null is invalid.
```
nsricardor: This adds PATCH-style field-presence semantics to a PUT representation. RFC 9110 defines PUT as replacing the target resource with the supplied representation. Omitted posixAcl and atimeUpdateIntervalSeconds instead silently retain server state, while explicit values replace it and null is rejected. This makes the update contract depend on hidden prior state and adds unnecessary absent/null/value complexity. This PATCH-style behavior has no precedent outside the File Storage endpoints. Please make these fields required on update and replace their values directly.

vinimdocarmo: I understand the PATCH-style situation. The design came from introducing those without breaking changes to the API. If we make the new fields required on PUT it will mean a breaking change: old clients calling this endpoint will get an error when currently they succeed.

[UNI specs Section 7.7](https://github.com/nscaledev/uni-specifications/blob/main/SPECIFICATION.md#77-api-versioning) says:
> Public-facing APIs must be backward compatible within a version. Additive changes — new optional fields, new endpoints, new enum values — are permitted without a version bump. Removing fields, changing field semantics, or altering response structure requires a version bump.

Adding them as optional and having deterministic replacement defaults would be RFC-correct, but it would risk existing settings (applied by hand in the backend) becoming the default values again.

I'm now thinking about the alternatives we have.

vinimdocarmo: To avoid breaking changes we've landed on:
PUT
posix acl and atime optional; when null or omitted: false and 0 by default


GET
additive: will return whatever is stored in the CR


POST
optional with false and 0 by default

pcriv (approved):

